PR #2, opened 2015-03-01
Ufl interfacing

dham:
Provide an interface enabling FInAT Kernels to be employed from PyOP2.

Also include:
*  a basic tensor product implementation
*  fixes for Jacobians, determinants, and integrals.
*  introduce our own Variable subclass


mlange05:
This looks good, merging.

PR #3, opened 2015-03-04
BindingMapper

mlange05:
This merge adds a simple mapper that inserts `ForAlls` for any unbound indices in a `Recipe`. The interpreter now invokes this mapper before evaluating a `Recipe` instead of adding implicit `ForAlls` during evaluation. The merge also includes various bugfixes.

### DIFF
--- a/finat/mappers.py
+++ b/finat/mappers.py
@@ -224,5 +224,14 @@ class BindingMapper(IdentityMapper):
         body = self.rec(expr.body)
         for idx in indices:
             self.bound_above.remove(idx)
-            self.bound_below.add(idx)
         return IndexSum(indices, body)
+
+    def map_for_all(self, expr):
+        indices = expr.indices
+        for idx in indices:
+            self.bound_above.add(idx)
+        body = self.rec(expr.body)
+        for idx in indices:
+            self.bound_above.remove(idx)
+            self.bound_below.add(idx)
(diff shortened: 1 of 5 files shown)


finat/mappers.py
```diff
@@ -188,3 +191,40 @@ def map_index_sum(self, expr, *args, **kwargs):
...
+class BindingMapper(IdentityMapper):
+    """A mapper that binds free indices in recipes using ForAlls."""
+
+    def __init__(self, kernel_data):
+        """
+        :arg context: a mapping from variable names to values
+        """
+        super(BindingMapper, self).__init__()
+        self.bound_above = {}
+        self.bound_below = {}
```
dham: Is there a reason to use dictionaries here rather than sets? As far as I can see you always set the values in the dictionary to None so you are effectively using the dictionary as a set of keys.

tests/test_interpreter.py
```diff
@@ -8,6 +8,7 @@ def i():
     return finat.indices.DimensionIndex(10)
 
 
+@pytest.mark.xfail
 def test_invalid_binding(i):
```
dham: This highlights a problem. When leaving an index sum then the index should not become bound below, it should become actually undefined (ie just not added to the bound_below set). This is a key difference between IndexSum (which is a reduction operation) and ForAll (which is not)

PR #5, opened 2015-03-25
Coffee RHS assembly

mlange05:
This merge adds the final pieces for RHS assembly via FInAT and Coffee through the Firedrake `finat_loop` interface. The generated kernel code is functional for Lagrange elements with geometry and `grad` derivatives, although it is not yet optimised and currently only supports RHS assembly due to double indirections in the kernel signatures.


dham:
Yes. This looks good!

PR #6, opened 2015-03-29
Index permuation

dham:
This branch deals with the fact that the order of the ForAll statements may not correspond with the order in which the indices occur in Recipes. Recipe objects now record the required generalised transpose and the interpreter honours this.

As part of this work, the BindingMapper is now stateless and passes binding information as arguments rather than as attributes of the mapper object.

There is an xfailing test for the coffee compiler which highlights that this fix has not been applied there yet. That xfailing test highlights a further issue with the coffee compiler's handing of multiple ForAll statements.

### DIFF
--- a/finat/__init__.py
+++ b/finat/__init__.py
@@ -6,6 +6,8 @@ from product_elements import ScalarProductElement
 from points import PointSet
 from utils import KernelData, Kernel
 from derivatives import div, grad, curl
+from indices import *
+from ast import *
 import interpreter
 import quadrature
 import ufl_interface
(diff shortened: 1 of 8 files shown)


mlange05:
Generally, these changes look good to me and I agree with making the BindingMapper stateless.

The one thing I am not sure about is passing a pre-computed index permutation directly to the respective interpreter. Instead, should the BindingMapper not "fix" the Recipe by hoisting the `ForAlls` into the order given by the Recipe signature? This would go along with other planned optimisations, eg. hoisting the `J` loops. 

I realise though that this requires some non-trivial implementation work, so I am happy for the current solution to be merged as is.

dham:
The reason the permutation is necessary is that the order of indices in the output array may not necessarily be the optimal ordering for the assembly operation. In particular, if you have loop carried dependencies (which happens in the Bernstein case), then you don't even get the right answer if you arbitrarily reorder the loop nest!

PR #8, opened 2016-04-25
Proposed fixes

miklos1:


finat/vectorfiniteelement.py
```diff
@@ -65,9 +65,11 @@ def basis_evaluation(self, q, entity=None, derivative=0):
         i = gem.Index(extent=self._dimension)
         vi = gem.Index(extent=self._dimension)
 
-        new_indices = indices[:qi] + i + indices[qi: -d] + vi + indices[-d:]
+        new_indices = indices[:qi] + (i,) + indices[qi: len(indices) - d] + (vi,) + indices[len(indices) - d:]
```
wence-: Is `indices[:-d]` somehow not the same as `indices[:len(indices)-d]`?

miklos1: They are different when d = 0.

wence-: Oh, ugh.

PR #9, opened 2016-09-13
Generalise VectorFiniteElement -> TensorFiniteElement

miklos1:
This is largely based on part of `tensor_product` branch, but rather than just adding a rank two tensor (AKA matrix) class, it adds a general tensor class of arbitrary rank, so the old `VectorFiniteElement` can be removed.


finat/tensorfiniteelement.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class TensorFiniteElement(FiniteElementBase):
+
+    def __init__(self, element, shape):
+        r"""A Finite element whose basis functions have the form:
+
+        .. math::
+
+            \boldsymbol\phi_{i \alpha \beta} = \mathbf{e}_{\alpha} \mathbf{e}_{\beta}^{\mathrm{T}}\phi_i
```
miklos1: This probably needs updating as we no longer assume rank = 2.

finat/tensorfiniteelement.py
```diff
@@ -0,0 +1,92 @@
...
+        qi = len(q.index_shape) + len(self._base_element.index_shape)
+        d = derivative
+
+        # New basis function and value indices.
+        i = tuple(gem.Index(extent=d) for d in self._shape)
+        vi = tuple(gem.Index(extent=d) for d in self._shape)
+
+        new_indices = indices[:qi] + i + indices[qi: len(indices) - d] + vi + indices[len(indices) - d:]
+
+        return gem.ComponentTensor(gem.Product(reduce(gem.Product,
```
miklos1: This is updated (compared to `tensor_product`) to handle arbitrary rank.

PR #12, opened 2016-10-19
Revise point sets and quadrature rules

miklos1:
This change redesigns point sets and quadrature rules. Quadrature rules now use point sets to store their points and augment them with weights. Finite elements no longer require quadrature rules for basis evaluation, but accept point sets instead.


finat/gauss_jacobi.py
```diff
@@ -1,115 +0,0 @@
...
-            pn1 = p
-        return p
-
-
-def eval_jacobi_deriv(a, b, n, x):
-    """Evaluates the first derivative of P_{n}^{a,b} at a point x."""
-    if n == 0:
-        return 0.0
-    else:
-        return 0.5 * (a + b + n + 1) * eval_jacobi(a + 1, b + 1, n - 1, x)
```
miklos1: This file duplicates code that is already in FIAT, thus removed.

finat/point_set.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class AbstractPointSet(with_metaclass(ABCMeta)):
+    """A way of specifying a known set of points, perhaps with some
+    (tensor) structure."""
+
+    @abstractproperty
+    def points(self):
+        """A flattened numpy array of points with shape
+        (# of points, point dimension)."""
```
miklos1: We use this to flatten the points for backwards compatibility with FIAT elements.

finat/point_set.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @abstractproperty
+    def indices(self):
+        """GEM indices with matching shape and extent to the structure of the
+        point set."""
+
+    @abstractproperty
+    def expression(self):
+        """GEM expression describing the points, with free indices
+        ``self.indices`` and shape (point dimension,)."""
```
miklos1: It seems that this is rarely used, only for `CellCoordinate` and `FacetCoordianate`.

finat/point_set.py
```diff
@@ -0,0 +1,123 @@
...
+    """A basic point set with no internal structure."""
+
+    def __init__(self, points):
+        points = numpy.asarray(points)
+        assert len(points.shape) == 2
+        self.points = points
+
+    @cached_property
+    def points(self):
+        pass  # set at initialisation
```
miklos1: Property needed to satisfy the abstract interface.

finat/point_set.py
```diff
@@ -0,0 +1,123 @@
...
+# class StroudPointSet(TensorPointSet, DuffyMappedMixin):
+#     """A set of points with the structure required for Stroud quadrature."""
+
+#     def __init__(self, factor_sets):
+#         super(StroudPointSet, self).__init__(factor_sets)
+
+
+# class GaussLobattoPointSet(PointSet):
+#     """A set of 1D Gauss Lobatto points. This is a separate class in order
+#     to allow elements to apply spectral element tricks."""
```
miklos1: I commented these out because they are currently not used, likely broken for not being updated to the new interface, but might be useful later.

finat/quadrature.py
```diff
@@ -1,150 +1,196 @@
...
 
+from finat.point_set import PointSet, TensorPointSet
 
-class QuadratureRule(object):
-    """Object representing a quadrature rule as a set of points
-    and a set of weights.
 
-    :param cell: The :class:`~FIAT.reference_element.ReferenceElement`
-    on which this quadrature rule is defined.
+def make_quadrature(ref_el, degree, scheme="default"):
```
miklos1: This mostly mimics `FIAT.quadrature_schemes.create_quadrature`. Preferred quadrature argument currently removed. I argue for a gradual approach where we first have sum factorisation with the potential need to manually specify the quadrature rule, and then we can improve the automatic quadrature scheme selection.

finat/quadrature.py
```diff
@@ -1,150 +1,196 @@
...
     """
+    Generate quadrature rule for given reference element
+    that will integrate an polynomial of order 'degree' exactly.
 
-    def __init__(self, cell, points, weights):
-        self.cell = cell
```
miklos1: Not convinced that we need a cell in quadrature rules.

finat/quadrature.py
```diff
@@ -1,150 +1,196 @@
...
+    def point_set(self):
+        return TensorPointSet(q.point_set for q in self.factors)
+
+    @cached_property
+    def weight_expression(self):
+        return reduce(gem.Product, (q.weight_expression for q in self.factors))
+
+    # def refactor(self, dims):
+    #     """Refactor this quadrature rule into a tuple of quadrature rules with
+    #     the dimensions specified."""
```
miklos1: Not sure what this was supposed to do, but we probably don't need this.

finat/quadrature.py
```diff
@@ -1,150 +1,196 @@
...
+#         weightset = tuple(lookup[c, p][1] for c, p, _ in cpq)
+
+#         if len(cpq) == 1:
+#             super(GaussLobattoQuadrature, self).__init__(
+#                 cell, pointset[0], weightset[0])
+#         else:
+#             super(GaussLobattoQuadrature, self).__init__(
+#                 cell,
+#                 TensorPointSet(pointset),
+#                 weightset)
```
miklos1: This was commented out for the same reason as in `point_set.py`: not currently used, not updated for the new interface, but might contain useful logic.

miklos1:
Merging this for the lack of objections. @dham can also complain later.

PR #13, opened 2016-10-20
Report if basis functions are cellwise constant

miklos1:
This change makes FInAT report if a basis function is cellwise constant, so TSFC can optimise evaluation. This is achieved through an **API change**:

`basis_evaluation` no longer returns point structure as shape, but rather as free indices. If the value of the basis functions does not depend on some point indices, then those indices simply won't appear in the free indices of the return expression.

### DIFF
--- a/finat/fiat_elements.py
+++ b/finat/fiat_elements.py
@@ -45,23 +45,20 @@ class FiatElementBase(FiniteElementBase):
         # Work out the correct transposition between FIAT storage and ours.
         tr = (2, 0, 1) if self.value_shape else (1, 0)
 
-        # Convert the FIAT tabulation into a gem tensor. Note that
-        # this does not exploit the symmetry of the derivative tensor.
         if derivative:
             e = np.eye(dim, dtype=np.int)
             tensor = np.empty((dim,) * derivative, dtype=np.object)
             it = np.nditer(tensor, flags=['multi_index', 'refs_ok'], op_flags=["writeonly"])
             while not it.finished:
                 derivative_multi_index = tuple(e[it.multi_index, :].sum(0))
-                it[0] = gem.Literal(restore_shape(fiat_tab[derivative_multi_index].transpose(tr), ps))
+                it[0] = gem.Indexed(gem.Literal(restore_shape(fiat_tab[derivative_multi_index].transpose(tr), ps)),
+                                    pi + i + vi)
                 it.iternext()
-            tensor = gem.ListTensor(tensor)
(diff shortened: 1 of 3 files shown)


finat/fiat_elements.py
```diff
@@ -32,36 +31,46 @@ def basis_evaluation(self, ps, entity=None, derivative=0):
...
 
-        fiat_tab = self._fiat_element.tabulate(derivative, ps.points, entity)
+        if derivative < self._degree:
+            points = ps.points
+            point_indices = ps.indices
+        elif derivative == self._degree:
+            # Tabulate on cell centre
+            points = np.mean(self._cell.get_vertices(), axis=0, keepdims=True)
+            entity = (self._cell.get_dimension(), 0)
+            point_indices = ()  # no point indices used
```
miklos1: Evaluation at cell centre when the specific derivative is cellwise constant.

finat/fiat_elements.py
```diff
@@ -32,36 +31,46 @@ def basis_evaluation(self, ps, entity=None, derivative=0):
...
+        else:
+            return gem.Zero(tuple(index.extent for index in i + vi + di))
```
miklos1: Return appropriately shaped zeros for high derivatives.

finat/fiat_elements.py
```diff
@@ -32,36 +31,46 @@ def basis_evaluation(self, ps, entity=None, derivative=0):
...
-            tensor = gem.ListTensor(tensor)
+            tensor = gem.Indexed(gem.ListTensor(tensor), di)
         else:
-            tensor = gem.Literal(restore_shape(fiat_tab[(0,) * dim].transpose(tr), ps))
+            tensor = tensor[()]
 
-        return gem.ComponentTensor(gem.Indexed(tensor,
-                                               di + pi + i + vi),
-                                   pi + i + vi + di)
+        return gem.ComponentTensor(tensor, i + vi + di)
```
wence-: OK, removing point indices from the returned CT, so they're free.

finat/fiat_elements.py
```diff
@@ -32,36 +31,46 @@ def basis_evaluation(self, ps, entity=None, derivative=0):
         :param entity: the cell entity on which to tabulate.
         :param derivative: the derivative to take of the basis functions.
         '''
-
```
miklos1: This function looks a bit ugly, but I'm not sure yet how to clean it up.

miklos1:
Merging this now, because the next pull request is coming up.

PR #14, opened 2016-10-21
Working implementation of the tensor product element

miklos1:
What the title says. I have written some explanation to the code in `TensorProductElement.basis_evaluation`. Please check if that makes sense.


finat/tensor_product.py
```diff
@@ -0,0 +1,177 @@
...
+
+class TensorProductElement(FiniteElementBase):
+
+    def __init__(self, factors):
+        self.factors = tuple(factors)
+        assert all(fe.value_shape == () for fe in self.factors)
+
+        self._cell = TensorProductCell(*[fe.cell for fe in self.factors])
+        # self._degree = sum(fe.degree for fe in factors)  # correct?
+        # self._degree = max(fe.degree for fe in factors)  # FIAT
```
miklos1: For now, I just want to be clear that it isn't used yet, and listed the existing conventions for a future cleanup.

finat/tensor_product.py
```diff
@@ -0,0 +1,177 @@
...
+            # contributions from the subelements.
+            scalars = []
+            for fe, ps_, e, D, alpha in zip(self.factors, ps_factors, entities, Ds, alphas):
+                # Ask the subelement to tabulate at the required derivative order.
+                value = fe.basis_evaluation(ps_, entity=e, derivative=sum(D))
+                # Nice, but now we have got a subelement derivative
+                # tensor of the given order, while we only need a
+                # specific derivative.  So we convert the subelement
+                # canonical multiindex to tensor multiindex.
+                d = tuple(chain(*[(dim,) * count for dim, count in enumerate(D)]))
```
wence-: So you have to do this dance between tensor and canonical multi index because it's easier to decide which derivatives apply to which sub element in the tensor case, but everything else wants the canonical case, right?

I think like the reader of this code's brain could be eased if you wrapped some of these iterable one-liners up in functions that say what they do: e.g. `d = to_tensor(D)` and earlier `D_ = to_canonical(delta)`...

finat/tensor_product.py
```diff
@@ -0,0 +1,177 @@
...
+                # Turn basis shape to free indices, select the right
+                # derivative entry, and collect the result.
+                scalars.append(gem.Indexed(value, alpha + d))
+            # Multiply the values from the subelements and insert the
+            # result into the derivative tensor.
+            tensor[delta] = reduce(gem.Product, scalars)
```
wence-: Oh, and you need the tensor multi index here too to populate the derivative tensor.

miklos1:
I'm now thinking that the implementation could be cleaner (and more efficient because of fewer recomputations) if we made the FInAT API more FIAT-like rather than what one would prefer for translating UFL/TSFC modified terminals. That is, instead of returning a derivative tensor for a given derivative order, get a jet order as an argument, and return map with all the canonical derivative multiindices mapping to GEM expressions. While that would require a conversion on the TSFC side (just as we have on master), but we could avoid this back and forth dancing when building FInAT elements from other FInAT elements.

miklos1:
Refactoring API before merging this...

PR #15, opened 2016-10-24
Make the basis_evaluation API more like FIAT

miklos1:
This is on top of #14, so it also gives a working tensor product implementation. The additional change is to make `basis_evaluation` receive an `order` argument instead of `derivative`, evaluating all derivatives up to `order`, and returning them in a `dict` that maps derivative multiindices to GEM expressions. This makes the implementation of `TensorProductElement.basis_evaluation` much more legible.


finat/fiat_elements.py
```diff
@@ -23,54 +25,40 @@ def index_shape(self):
...
-        if derivative:
-            tensor = gem.Indexed(gem.ListTensor(tensor), di)
-        else:
-            tensor = tensor[()]
-
-        return gem.ComponentTensor(tensor, i + vi + di)
+        fiat_result = self._fiat_element.tabulate(order, ps.points, entity)
+        result = {}
+        for alpha, table in iteritems(fiat_result):
+            # Points be the first dimension, not last.
```
wence-: Gone a bit Westcountry here.

finat/fiat_elements.py
```diff
@@ -23,54 +25,40 @@ def index_shape(self):
...
+            reordering = list(range(len(table.shape)))
+            reordering = [reordering[-1]] + reordering[:-1]
```
wence-: If only there were a `rotatef`.

finat/fiat_elements.py
```diff
@@ -23,54 +25,40 @@ def index_shape(self):
...
 
-    def basis_evaluation(self, ps, entity=None, derivative=0):
+    def basis_evaluation(self, order, ps, entity=None):
         '''Return code for evaluating the element at known points on the
         reference element.
 
         :param ps: the point set.
         :param entity: the cell entity on which to tabulate.
         :param derivative: the derivative to take of the basis functions.
         '''
```
dham: Please to be fixing docstrings to match interface changes.

PR #22, opened 2017-01-12
Fix element wrappers for GL and GLL

miklos1:


finat/fiat_elements.py
```diff
@@ -87,6 +87,11 @@ def basis_evaluation(self, order, ps, entity=None):
         return result
 
 
+class Regge(FiatElementBase):  # naturally tensor valued
```
miklos1: Regge is not a scalar element, but is naturally tensor valued.

PR #23, opened 2017-02-11
FIAT-compatible API for node allocation

### LINKED ISSUE #20: Provide API for node allocation
Currently, Firedrake function spaces allocate nodes based on the entity DoFs provided by FIAT elements, while TSFC implements the elements using FInAT. It would be nice to use the same elements for node allocation _and_ for form compilation.

miklos1:
Resolves #20.

### DIFF
--- a/finat/finiteelementbase.py
+++ b/finat/finiteelementbase.py
@@ -1,11 +1,16 @@
 from __future__ import absolute_import, print_function, division
-from six import with_metaclass, iteritems
+from six import with_metaclass, iteritems, itervalues
 
 from abc import ABCMeta, abstractproperty, abstractmethod
 from itertools import chain
 
+import numpy
+
 import gem
+from gem.optimise import aggressive_unroll
 from gem.utils import cached_property
+
+from finat.quadrature import make_quadrature
 
 
 class FiniteElementBase(with_metaclass(ABCMeta)):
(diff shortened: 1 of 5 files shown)


finat/finiteelementbase.py
```diff
@@ -51,3 +82,44 @@ def basis_evaluation(self, order, ps, entity=None):
...
+        # Tabulate basis functions on the facet
+        vals, = itervalues(elem.basis_evaluation(0, quad.point_set, entity=(entity_dim, f)))
+        # Integrate the square of the basis functions on the facet.
+        ints = gem.IndexSum(
+            gem.Product(gem.IndexSum(gem.Product(gem.Indexed(vals, beta + zeta),
+                                                 gem.Indexed(vals, beta + zeta)), zeta),
+                        quad.weight_expression),
+            quad.point_set.indices
+        )
+        ints = aggressive_unroll(gem.ComponentTensor(ints, beta)).array.flatten()
```
wence-: This works because gem constant folds, right?

finat/tensor_product.py
```diff
@@ -29,7 +29,35 @@ def cell(self):
...
-        raise NotImplementedError("Unused property.")
+        return tuple(fe.degree for fe in self.factors)
+
+    @cached_property
+    def _entity_dofs(self):
+        shape = tuple(fe.space_dimension() for fe in self.factors)
+        entity_dofs = {}
+        for dim in product(*[fe.cell.get_topology().keys()
+                             for fe in self.factors]):
+            entity_dofs[dim] = {}
```
wence-: You use `entity_dofs[dim]` as a temporary dict here.

finat/tensor_product.py
```diff
@@ -29,7 +29,35 @@ def cell(self):
...
+            for tuple_ei in product(*[sorted(topd) for topd in topds]):
+                tuple_vs = list(product(*[topd[ei]
+                                          for topd, ei in zip(topds, tuple_ei)]))
+                if tuple_vs:
+                    vs = list(numpy.ravel_multi_index(numpy.transpose(tuple_vs), shape))
+                else:
+                    vs = []
+                entity_dofs[dim][tuple_ei] = vs
+            # flatten entity numbers
+            entity_dofs[dim] = dict(enumerate(entity_dofs[dim][key]
```
wence-: And then overwrite here.  Maybe it would be clearer if these were two different variables?

finat/finiteelementbase.py
miklos1: Yes.

finat/tensor_product.py
miklos1: This is mostly based on the tensor product calculation of cell topology in `FIAT.reference_element`, but I agree, it is clearer to just give them different names.

PR #25, opened 2017-04-07
Add QuadratureElement

miklos1:
Similar to `QuadratureElement` in FFC, but this one returns symbolic identity matrices (GEM `Delta` nodes) rather than numerical ones, which makes it much easier to simplify away coefficient and maybe even argument evaluations.

### DIFF
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, print_function, division
 from six import with_metaclass
-from six.moves import range
+from six.moves import range, zip
 
 from abc import ABCMeta, abstractproperty
 from itertools import chain, product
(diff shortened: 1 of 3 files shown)


finat/point_set.py
```diff
@@ -80,3 +86,9 @@ def expression(self):
             for i in range(point_set.dimension):
                 result.append(gem.Indexed(point_set.expression, (i,)))
         return gem.ListTensor(result)
+
+    def almost_equal(self, other, tolerance=1e-12):
+        """Approximate numerical equality of point sets"""
+        return type(self) == type(other) and \
+            len(self.factors) == len(other.factors) and \
+            all(s.almost_equal(o) for s, o in zip(self.factors, other.factors))
```
miklos1: Forgot to propagate `tolerance`...

PR #27, opened 2017-05-03
Esd method

wence-:
Make entity_support_dofs a method of FiniteElementBase, rather than a free function (keep that for backwards compatibility).

### DIFF
--- a/finat/finiteelementbase.py
+++ b/finat/finiteelementbase.py
@@ -46,6 +46,40 @@ class FiniteElementBase(with_metaclass(ABCMeta)):
         element.'''
         return self._entity_closure_dofs
 
+    @cached_property
+    def _entity_support_dofs(self):
+        esd = {}
+        for entity_dim in self.cell.sub_entities.keys():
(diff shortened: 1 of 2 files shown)


miklos1:
So far the difference does not seem significant. Do you plan to have specific versions of `entity_support_dofs` for different finite element classes, or what is the reason for the change?

wence-:
Note that you now get the `entity_support_dofs` for all entities, rather than the specified entity (mirroring `entity_closure_dofs`).  I need all of them for the extruded layers stuff.  I guess I could just make the entity optional in the free function instead.  But it has always seemed strange to me how `e_c_d` is a method, but `e_s_d` is not.

miklos1:
OK, it used to be a method at some point in the past.

wence-:
Although this is ok, it runs into problems for enriched elements in 3D.  Since no-one really uses `entity_support_dofs`, and I only want this for "completeness" in extrusion, I will drop this for now.

PR #31, opened 2017-05-18
Cellwise constantness aware code generation for the point evaluation of Ciarlet elements

### LINKED ISSUE #30: Point evaluation fails to recognise cellwise constantness
Point evaluation takes a GEM expression describing the reference coordinates. This expression may have free indices, so that the finite element can be evaluated at a set of points. However, the quantity being evaluated may be cellwise constant, in this case it would neat to get an expression as a result that is independent of the specified reference point. However, this does not currently happen with structured point evaluation for Ciarlet elements. For example, the coefficient matrix for one derivative of a P1 triangular element looks like this:
```c
static const double t0[3][3] = {
    {-1.41421356237309, 0.0, 0.0}, 
    {1.41421356237309, 0.0, 0.0}, 
    {0.0, 0.0, 0.0}
};
```
The first polynomial in the Dubiner basis is constant, the other two depend on the reference coordinates. But those values, of course, don't actually matter, and as we can see they correspond to zero columns in the coefficient matrix.

miklos1:
Resolves #30. Adds the first ever unit test in FInAT.

### DIFF
--- a/finat/fiat_elements.py
+++ b/finat/fiat_elements.py
@@ -190,6 +190,16 @@ def point_evaluation_ciarlet(fiat_element, order, refcoords, entity):
     base_values = poly_set.get_expansion_set().tabulate(degree, [X])
     m = len(base_values)
     assert base_values.shape == (m, 1)
+    base_values_sympy = np.array(list(base_values.flat))
+
+    # Find constant polynomials
+    def is_const(expr):
+        try:
+            float(expr)
+            return True
+        except TypeError:
+            return False
+    const_mask = np.array(list(map(is_const, base_values_sympy)))
 
     # Convert SymPy expression to GEM
     mapper = gem.node.Memoizer(sympy2gem)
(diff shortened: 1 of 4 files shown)


finat/fiat_elements.py
```diff
@@ -205,16 +215,21 @@ def point_evaluation_ciarlet(fiat_element, order, refcoords, entity):
...
-            k = gem.Index()
-            result[alpha] = gem.ComponentTensor(
-                gem.IndexSum(
-                    gem.Product(gem.Indexed(gem.Literal(table), beta + (k,)),
-                                gem.Indexed(base_values, (k,))),
-                    (k,)
-                ),
-                beta
-            )
+            zerocols = np.isclose(abs(table).max(axis=tuple(range(table.ndim - 1))), 0.0)
```
wence-: Is the `abs` necessary?

finat/fiat_elements.py
```diff
@@ -205,16 +215,21 @@ def point_evaluation_ciarlet(fiat_element, order, refcoords, entity):
...
+            if all(np.logical_or(const_mask, zerocols)):
+                vals = base_values_sympy[const_mask]
+                result[alpha] = gem.Literal(table[..., const_mask].dot(vals))
```
wence-: There's an implicit conversion to float in the dot?

requirements-ext.txt
```diff
@@ -1 +1,2 @@
 numpy>=1.6.1
+singledispatch
```
wence-: This was previously a requirement, except we always installed in the context where single dispatch was already available I guess?

finat/fiat_elements.py
miklos1: Yes, because `max(-1.2, 0.01) == 0.01` so it would otherwise fail to check closeness to zero for negative numbers.

wence-: Oh yes, I missed the max part.

finat/fiat_elements.py
miklos1: I don't think the dot converts to float, but `gem.Literal` definitely casts to numpy array of floats.

requirements-ext.txt
miklos1: Yes, but it now came up as missing in Travis CI, since for the first time tests are added.

PR #32, opened 2017-07-17
Faster entity_support_dofs

miklos1:
Use GEM interpreter instead of `gem.optimise.aggressive_unroll`.

Requires firedrakeproject/tsfc#129.

Resolves firedrakeproject/firedrake#1098.

### DIFF
--- a/finat/finiteelementbase.py
+++ b/finat/finiteelementbase.py
@@ -7,7 +7,7 @@ from itertools import chain
 import numpy
 
 import gem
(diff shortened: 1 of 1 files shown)


miklos1:
The interpreter has a better representation for fast calculations than the symbolic constant folder. It was somewhat faster at first, but then after optimising the interpreter a bit (move loops from Python to inside numpy) it became significantly faster for that one example I tested it with.

Issue #19, opened 2016-12-12
Implement structured EnrichedElement and MixedElement


miklos1:
For the sum factorisation of RTCF, RTCE, NCF, NCE elements and of the appropriate H(div) and H(curl) elements on triangular prism cells.

miklos1:
Created a separate issue for `MixedElement` (#35). `EnrichedElement` was implemented in #33 and firedrakeproject/tsfc#132.

PR #33, opened 2017-07-20
Enriched element

miklos1:
The FInAT version of `FIAT.EnrichedElement`. Related to #19.

Requires firedrakeproject/tsfc#131.

### DIFF
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -1,0 +1,103 @@
+from __future__ import absolute_import, print_function, division
+from six.moves import map, zip
+
+import gem
+from gem.utils import cached_property
+
+from finat.finiteelementbase import FiniteElementBase
+
+
+class EnrichedElement(FiniteElementBase):
+    """A finite element whose basis functions are the union of the
+    basis functions of several other finite elements."""
+
+    def __init__(self, elements):
+        super(EnrichedElement, self).__init__()
+        self.elements = tuple(elements)
+
+    @cached_property
+    def cell(self):
+        result, = set(elem.cell for elem in self.elements)
+        return result
+
+    @cached_property
+    def degree(self):
+        return tree_map(max, *[elem.degree for elem in self.elements])
+
+    def entity_dofs(self):
+        '''Return the map of topological entities to degrees of
+        freedom for the finite element.'''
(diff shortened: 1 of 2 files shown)


finat/enriched.py
```diff
@@ -0,0 +1,103 @@
...
+        results = [element.point_evaluation(order, refcoords, entity)
+                   for element in self.elements]
+        return self._compose_evaluations(results)
+
+
+def tree_map(f, *args):
+    """Like the built-in :py:func:`map`, but applies to a tuple tree."""
+    nonleaf, = set(isinstance(arg, tuple) for arg in args)
+    if nonleaf:
+        ndim, = set(map(len, args))
```
wence-: assigned to but never used?

finat/enriched.py
```diff
@@ -0,0 +1,103 @@
...
+        def merge(tables):
+            tables = tuple(tables)
+            zeta = self.get_value_indices()
+            tensors = []
+            for elem, table in zip(self.elements, tables):
+                beta_i = elem.get_indices()
+                tensors.append(gem.ComponentTensor(
+                    gem.Indexed(table, beta_i + zeta),
+                    beta_i
+                ))
```
wence-: So each sub element provides a `CT(table[beta_i, zeta], beta_i)`, which has free indices zeta (the indices of the outer element)

finat/enriched.py
```diff
@@ -0,0 +1,103 @@
...
+            beta = self.get_indices()
+            return gem.ComponentTensor(
+                gem.Indexed(gem.Concatenate(*tensors), beta),
+                beta + zeta
+            )
```
wence-: And then you concatenate the sub element tables together and bind to the outer indices.

finat/enriched.py
miklos1: I only want the side effect actually, that is, a failure when arguments have different length. Perhaps
```python
assert len(set(map(len, args))) == 1
```
communicates the intention better, though I find the `len(set(map(len` part a bit inconvenient to read.

wence-: I see, the latter is clearer I think (although I agree with the readability), and less likely to be removed at a later date.

PR #37, opened 2017-07-31
Mixed element and "transposable" tensor element

miklos1:
Add a structured mixed element implemented through `EnrichedElement`, and make `TensorFiniteElement` "transposable": vector/tensor indices may come after or before the scalar basis function indices.

### DIFF
--- a/finat/mixed.py
+++ b/finat/mixed.py
@@ -1,0 +1,97 @@
+from __future__ import absolute_import, print_function, division
+from six import iteritems
+from six.moves import zip
+
+import numpy
+
+import gem
+
+from finat.finiteelementbase import FiniteElementBase
+from finat.enriched import EnrichedElement
+
+
+def MixedElement(elements):
+    """Constructor function for FEniCS-style mixed elements.
+
+    Implements mixed element using :py:class:`EnrichedElement` and
+    value shape transformations with :py:class:`MixedSubElement`.
+    """
+    sizes = [numpy.prod(element.value_shape, dtype=int)
+             for element in elements]
+    offsets = [int(offset) for offset in numpy.cumsum([0] + sizes)]
+    total_size = offsets.pop()
+    return EnrichedElement([MixedSubElement(element, total_size, offset)
+                            for offset, element in zip(offsets, elements)])
+
+
+class MixedSubElement(FiniteElementBase):
+    """Element wrapper that flattens value shape and places the flattened
+    vector in a longer vector of zeros."""
+
+    def __init__(self, element, size, offset):
+        assert 0 <= offset <= size
+        assert offset + numpy.prod(element.value_shape, dtype=int) <= size
+
+        super(MixedSubElement, self).__init__()
+        self.element = element
+        self.size = size
+        self.offset = offset
+
+    @property
+    def cell(self):
+        return self.element.cell
+
+    @property
+    def degree(self):
+        return self.element.degree
+
+    @property
+    def formdegree(self):
+        return self.element.formdegree
+
+    def entity_dofs(self):
+        return self.element.entity_dofs()
+
+    def entity_closure_dofs(self):
+        return self.element.entity_closure_dofs()
+
+    def space_dimension(self):
+        return self.element.space_dimension()
+
+    @property
+    def index_shape(self):
+        return self.element.index_shape
+
+    @property
+    def value_shape(self):
+        return (self.size,)
+
+    def _transform(self, v):
+        u = [gem.Zero()] * self.size
+        for j, zeta in enumerate(numpy.ndindex(self.element.value_shape)):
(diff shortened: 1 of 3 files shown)


finat/tensorfiniteelement.py
```diff
@@ -12,7 +12,7 @@
 
 class TensorFiniteElement(FiniteElementBase):
 
-    def __init__(self, element, shape):
+    def __init__(self, element, shape, transpose=False):
```
miklos1: Is this a good name?

finat/mixed.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Implements mixed element using :py:class:`EnrichedElement` and
+    value shape transformations with :py:class:`MixedSubElement`.
+    """
+    sizes = [numpy.prod(element.value_shape, dtype=int)
+             for element in elements]
+    offsets = [int(offset) for offset in numpy.cumsum([0] + sizes)]
+    total_size = offsets.pop()
+    return EnrichedElement([MixedSubElement(element, total_size, offset)
+                            for offset, element in zip(offsets, elements)])
```
wence-: OK, so you're using the same technology to pick these apart as already exists, and the magic is how you make the tabulation of the sub elements.

finat/mixed.py
```diff
@@ -0,0 +1,97 @@
...
+
+def MixedElement(elements):
+    """Constructor function for FEniCS-style mixed elements.
+
+    Implements mixed element using :py:class:`EnrichedElement` and
+    value shape transformations with :py:class:`MixedSubElement`.
+    """
+    sizes = [numpy.prod(element.value_shape, dtype=int)
+             for element in elements]
+    offsets = [int(offset) for offset in numpy.cumsum([0] + sizes)]
```
wence-: The `int` cast is because you don't want numpy int64 to sneak in?

finat/mixed.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @property
+    def value_shape(self):
+        return (self.size,)
+
+    def _transform(self, v):
+        u = [gem.Zero()] * self.size
+        for j, zeta in enumerate(numpy.ndindex(self.element.value_shape)):
+            u[self.offset + j] = gem.Indexed(v, zeta)
+        return u
```
wence-: OK so this puts the tabulation in the correct slice.

finat/mixed.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _transform_evaluation(self, core_eval):
+        beta = self.get_indices()
+        zeta = self.get_value_indices()
+
+        def promote(table):
+            v = gem.partial_indexed(table, beta)
+            u = gem.ListTensor(self._transform(v))
+            return gem.ComponentTensor(gem.Indexed(u, zeta), beta + zeta)
```
wence-: Promote table to appropriate shape.

finat/tensorfiniteelement.py
wence-: I think I am more confused by this part because I'm not following which indices are being transposed (and why it appears that not all of the relevant bits need to be)

miklos1: To put it plainly, should the DoF ordering be XXXX YYYY ZZZZ or XYZ XYZ XYZ XYZ (example for a P1 tetrahedron). `transpose=False` means the latter, `transpose=True` means the former. For normal vector elements, we use the latter ordering, FEniCS uses the former (expect for coordinates where FEniCS goes our way).

finat/tensorfiniteelement.py
```diff
@@ -63,15 +66,18 @@ def entity_dofs(self):
...
-        return self._base_element.index_shape + self._shape
+        if self._transpose:
+            return self._shape + self._base_element.index_shape
+        else:
+            return self._base_element.index_shape + self._shape
 
     @property
     def value_shape(self):
-        return self._base_element.value_shape + self._shape
+        return self._shape + self._base_element.value_shape
```
miklos1: This is relevant in the rare case when one takes the vector of naturally vector-valued element, e.g. `VectorElement('BDM', triangle, 2)`. In this case UFL defines what the value shape should be like, and UFL says:

    In [7]: VectorElement('BDM', triangle, 2, dim=3).value_shape()
    Out[7]: (3, 2)
    
    In [8]: VectorElement('BDM', triangle, 2, dim=3).reference_value_shape()
    Out[8]: (3, 2)

This is the same, irrespective of the DoF ordering.

finat/mixed.py
miklos1: Yes, I am adopting this obsessive casting, because in Python 3 `numpy.int64` is not instance of `int`, which sometimes cause annoyances.

finat/tensorfiniteelement.py
wence-: OK, I see.

PR #38, opened 2017-08-03
Esd method

wence-:
Make the interface for entity_support_dofs the same as that for entity_closure_dofs (i.e. that we can ask the element directly).


finat/finiteelementbase.py
```diff
@@ -50,6 +50,43 @@ def entity_closure_dofs(self):
         element.'''
         return self._entity_closure_dofs
 
+    @cached_property
+    def _entity_support_dofs(self):
+        esd = {}
+        for entity_dim in self.cell.sub_entities.keys():
```
miklos1: @dham: This is the problem.

PR #39, opened 2017-08-10
Add wrappers for more FIAT elements

miklos1:
Also adds a FInAT implementation of `ufl.BrokenElement`, similar to `FIAT.DiscontinuousElement`. So one can have, for example, a structured broken RTCF element.

### DIFF
--- a/finat/fiat_elements.py
+++ b/finat/fiat_elements.py
@@ -14,11 +14,11 @@ from finat.finiteelementbase import FiniteElementBase
 from finat.sympy2gem import sympy2gem
 
 
-class FiatElementBase(FiniteElementBase):
+class FiatElement(FiniteElementBase):
     """Base class for finite elements for which the tabulation is provided
     by FIAT."""
     def __init__(self, fiat_element):
-        super(FiatElementBase, self).__init__()
+        super(FiatElement, self).__init__()
         self._element = fiat_element
 
     @property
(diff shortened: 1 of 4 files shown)


finat/fiat_elements.py
```diff
@@ -278,11 +298,6 @@ def __init__(self, cell, degree):
         super(RaviartThomas, self).__init__(FIAT.RaviartThomas(cell, degree))
 
 
-class DiscontinuousRaviartThomas(VectorFiatElement):
-    def __init__(self, cell, degree):
-        super(DiscontinuousRaviartThomas, self).__init__(FIAT.DiscontinuousRaviartThomas(cell, degree))
```
miklos1: This is virtually the same as `DiscontinuousElement` + `RaviartThomas`, so not need.

finat/trace.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import absolute_import, print_function, division
+
+import FIAT
+
+from finat.fiat_elements import ScalarFiatElement
+
+
+class HDivTrace(ScalarFiatElement):
+    def __init__(self, cell, degree):
+        super(HDivTrace, self).__init__(FIAT.HDivTrace(cell, degree))
```
miklos1: @thomasgibson could replace this FIAT wrapper at some point with an actual FInAT implementation.

finat/fiat_elements.py
miklos1: Due to the removal of this, this pull request must be merged at the same as firedrakeproject/tsfc#147.

PR #44, opened 2017-10-20
Fix excessive memory consumption in entity_support_dofs: Part 2

miklos1:
I need this for more elements in order to run high-order H(div) and H(curl) elements.

### DIFF
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -1,5 +1,11 @@
 from __future__ import absolute_import, print_function, division
+from six import iteritems
 from six.moves import map, zip
+
+from functools import partial
+from operator import add, methodcaller
+
+import numpy
 
 import gem
 from gem.utils import cached_property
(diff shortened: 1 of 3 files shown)


finat/enriched.py
```diff
@@ -118,3 +129,24 @@ def tree_map(f, *args):
...
+    """
+    entity_dofs = {dim: {i: [] for i in entities}
+                   for dim, entities in iteritems(ref_el.get_topology())}
+    offsets = numpy.cumsum([0] + list(e.space_dimension()
+                                      for e in elements), dtype=int)
+    for i, d in enumerate(map(method, elements)):
+        for dim, dofs in iteritems(d):
+            for ent, off in iteritems(dofs):
+                entity_dofs[dim][ent] += list(map(partial(add, offsets[i]), off))
+    return entity_dofs
```
miklos1: This function is pulled from FIAT and slightly generalised.

finat/enriched.py
```diff
@@ -118,3 +129,24 @@ def tree_map(f, *args):
...
+    :returns: concatenated entity DoFs dict
+    """
+    entity_dofs = {dim: {i: [] for i in entities}
+                   for dim, entities in iteritems(ref_el.get_topology())}
+    offsets = numpy.cumsum([0] + list(e.space_dimension()
+                                      for e in elements), dtype=int)
+    for i, d in enumerate(map(method, elements)):
+        for dim, dofs in iteritems(d):
+            for ent, off in iteritems(dofs):
+                entity_dofs[dim][ent] += list(map(partial(add, offsets[i]), off))
```
wence-: These come out sorted because the sub entities are sorted, and enriched doesn't interleave.

PR #50, opened 2019-02-16
Cyrus finat branch

cyruscycheng21:


finat/fiat_elements.py
```diff
@@ -292,6 +292,10 @@ class DiscontinuousLagrange(ScalarFiatElement):
     def __init__(self, cell, degree):
         super(DiscontinuousLagrange, self).__init__(FIAT.DiscontinuousLagrange(cell, degree))
 
+class DPC(ScalarFiatElement):
```
thomasgibson: Should be another line (2 lines total) between class definitions

PR #54, opened 2019-06-28
Mimetic Spectral Elements

celdred:
This pull request adds mimetic spectral elements (described at https://www.sciencedirect.com/science/article/pii/S0021999113006414 and https://arxiv.org/abs/1111.4304), both the primal and dual complexes. This is done by adding the relevant 1D H1 element for the dual complex: Extended Gauss Legendre (EGL); and the L2 edge elements associated with GLL and EGL. These are non-Ciarlet finite elements that use a basis that histopolates (http://people.math.sfu.ca/~nrobidou/public_html/prints/histogram/histogram.pdf) rather than interpolates. The nD deRham complex on hypercubes is then accessible using the variant keyword (either “mse” or “dualmse”) with the relevant elements (DQ, Q, RTCF, etc.).

It requires https://github.com/FEniCS/ufl/pull/5 and https://github.com/FEniCS/fiat/pull/29.

It also requires correct (integral-preserving) L2 pullbacks, which are implemented at https://github.com/FEniCS/ufl/pull/4.


wence-:
This could be merged if we got the FIAT bit in, the UFL changes are done.

PR #56, opened 2020-06-10
Variants for BDM/RT and Nedelec spaces.

wence-:


wence-:
FIAT side is in, so this is safe to merge.

PR #57, opened 2020-06-18
Dual Evaluation

mkan88:
Performing dual evaluation in FInAT has been a long awaited feature in firedrake. The code is adapted from [TSFC #206](https://github.com/firedrakeproject/tsfc/pull/206) To be used with corresponding branches in [TSFC #218](https://github.com/firedrakeproject/tsfc/pull/218) and [Firedrake #1743](https://github.com/firedrakeproject/firedrake/pull/1743).

### DIFF
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -127,11 +127,12 @@ class EnrichedElement(FiniteElementBase):
                    for element in self.elements]
         return self._compose_evaluations(results)
 
+    @property
     def dual_basis(self):
         from itertools import chain
 
         # Note: NOT TESTED!!!
-        elements_dual_basis = [element.dual_basis() for element in self.elements]
+        elements_dual_basis = [element.dual_basis for element in self.elements]
         return tuple(chain(*elements_dual_basis))
 
     @property
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -128,24 +128,26 @@ class EnrichedElement(FiniteElementBase):
         return self._compose_evaluations(results)
 
     def dual_basis(self):
-        from itertools import chain, zip_longest
+        from itertools import chain
 
+        # Note: NOT TESTED!!!
         elements_dual_basis = [element.dual_basis() for element in self.elements]
+        return tuple(chain(*elements_dual_basis))
 
-        enriched_dual_basis = []
-        for elements_dual in zip_longest(*elements_dual_basis, fill_value=tuple()):
-            enriched_derivs = []
-            for elements_derivs in zip_longest(*elements_dual, fill_value=tuple()):
-                enriched_pts_in_derivs = []
-                for elements_tups in zip_longest(*elements_derivs, fill_value=tuple()):
-                    # TODO: Combine repeated points?
-                    enriched_pts_in_derivs.extend(elements_tups)
-                # Leave one empty tup for safety?
-                for _ in range(enriched_pts_in_derivs.count(tuple())-2):
-                    enriched_pts_in_derivs.remove(tuple())
-                enriched_derivs.append(tuple(enriched_pts_in_derivs))
-            enriched_dual_basis.append(tuple(enriched_derivs))
-        return tuple(enriched_dual_basis)
+        # enriched_dual_basis = []
+        # for elements_dual in zip_longest(*elements_dual_basis, fill_value=tuple()):
+        #     enriched_derivs = []
+        #     for elements_derivs in zip_longest(*elements_dual, fill_value=tuple()):
+        #         enriched_pts_in_derivs = []
+        #         for elements_tups in zip_longest(*elements_derivs, fill_value=tuple()):
+        #             # TODO: Combine repeated points?
+        #             enriched_pts_in_derivs.extend(elements_tups)
+        #         # Leave one empty tup for safety?
+        #         for _ in range(enriched_pts_in_derivs.count(tuple())-2):
+        #             enriched_pts_in_derivs.remove(tuple())
+        #         enriched_derivs.append(tuple(enriched_pts_in_derivs))
+        #     enriched_dual_basis.append(tuple(enriched_derivs))
+        # return tuple(enriched_dual_basis)
 
     @property
     def mapping(self):
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -128,26 +128,28 @@ class EnrichedElement(FiniteElementBase):
         return self._compose_evaluations(results)
 
     def dual_basis(self):
-        from itertools import chain
+        from itertools import chain, zip_longest
 
-        # Note: NOT TESTED!!!
         elements_dual_basis = [element.dual_basis() for element in self.elements]
-        return tuple(chain(*elements_dual_basis))
 
-        # enriched_dual_basis = []
-        # for elements_dual in zip_longest(*elements_dual_basis, fill_value=tuple()):
-        #     enriched_derivs = []
-        #     for elements_derivs in zip_longest(*elements_dual, fill_value=tuple()):
-        #         enriched_pts_in_derivs = []
-        #         for elements_tups in zip_longest(*elements_derivs, fill_value=tuple()):
-        #             # TODO: Combine repeated points?
-        #             enriched_pts_in_derivs.extend(elements_tups)
-        #         # Leave one empty tup for safety?
-        #         for _ in range(enriched_pts_in_derivs.count(tuple())-2):
-        #             enriched_pts_in_derivs.remove(tuple())
-        #         enriched_derivs.append(tuple(enriched_pts_in_derivs))
-        #     enriched_dual_basis.append(tuple(enriched_derivs))
-        # return tuple(enriched_dual_basis)
+        enriched_dual_basis = []
+        for elements_dual in zip_longest(*elements_dual_basis, fill_value=[]):
+            enriched_derivs = []
+            for elements_deriv in zip_longest(*elements_dual, fill_value=[]):
+                enriched_pts_in_derivs = []
+                for elements_tups in zip_longest(*elements_deriv, fill_value=[]):
+                    # TODO: Combine repeated points?
+                    # TODO: Remove empty tuples?
+                    for tups in elements_tups:
+                        try:
+                            elements_point_set, weight_tensor, alpha_tensor = elements_tups
+                        except ValueError:  # Empty
+                            continue
+
+                    enriched_pts_in_derivs.append(chain(*elements_tups))
+                enriched_derivs.append(tuple(enriched_pts_in_derivs))
+            enriched_dual_basis.append(tuple(enriched_derivs))
+        return tuple(enriched_dual_basis)
 
     @property
     def mapping(self):
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -133,20 +133,16 @@ class EnrichedElement(FiniteElementBase):
         elements_dual_basis = [element.dual_basis() for element in self.elements]
 
         enriched_dual_basis = []
-        for elements_dual in zip_longest(*elements_dual_basis, fill_value=[]):
+        for elements_dual in zip_longest(*elements_dual_basis, fill_value=tuple()):
             enriched_derivs = []
-            for elements_deriv in zip_longest(*elements_dual, fill_value=[]):
+            for elements_derivs in zip_longest(*elements_dual, fill_value=tuple()):
                 enriched_pts_in_derivs = []
-                for elements_tups in zip_longest(*elements_deriv, fill_value=[]):
+                for elements_tups in zip_longest(*elements_derivs, fill_value=tuple()):
                     # TODO: Combine repeated points?
-                    # TODO: Remove empty tuples?
-                    for tups in elements_tups:
-                        try:
-                            elements_point_set, weight_tensor, alpha_tensor = elements_tups
-                        except ValueError:  # Empty
-                            continue
-
-                    enriched_pts_in_derivs.append(chain(*elements_tups))
+                    enriched_pts_in_derivs.extend(elements_tups)
+                # Leave one empty tup for safety?
+                for _ in range(enriched_pts_in_derivs.count(tuple())-2):
+                    enriched_pts_in_derivs.remove(tuple())
                 enriched_derivs.append(tuple(enriched_pts_in_derivs))
             enriched_dual_basis.append(tuple(enriched_derivs))
         return tuple(enriched_dual_basis)
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -134,21 +134,6 @@ class EnrichedElement(FiniteElementBase):
         elements_dual_basis = [element.dual_basis() for element in self.elements]
         return tuple(chain(*elements_dual_basis))
 
-        # enriched_dual_basis = []
-        # for elements_dual in zip_longest(*elements_dual_basis, fill_value=tuple()):
-        #     enriched_derivs = []
-        #     for elements_derivs in zip_longest(*elements_dual, fill_value=tuple()):
-        #         enriched_pts_in_derivs = []
-        #         for elements_tups in zip_longest(*elements_derivs, fill_value=tuple()):
-        #             # TODO: Combine repeated points?
-        #             enriched_pts_in_derivs.extend(elements_tups)
-        #         # Leave one empty tup for safety?
-        #         for _ in range(enriched_pts_in_derivs.count(tuple())-2):
-        #             enriched_pts_in_derivs.remove(tuple())
-        #         enriched_derivs.append(tuple(enriched_pts_in_derivs))
-        #     enriched_dual_basis.append(tuple(enriched_derivs))
-        # return tuple(enriched_dual_basis)
-
     @property
     def mapping(self):
         mappings = set(elem.mapping for elem in self.elements)
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -127,6 +127,30 @@ class EnrichedElement(FiniteElementBase):
                    for element in self.elements]
         return self._compose_evaluations(results)
 
+    def dual_basis(self):
+        from itertools import chain, zip_longest
+
+        elements_dual_basis = [element.dual_basis() for element in self.elements]
+
+        enriched_dual_basis = []
+        for elements_dual in zip_longest(*elements_dual_basis, fill_value=[]):
+            enriched_derivs = []
+            for elements_deriv in zip_longest(*elements_dual, fill_value=[]):
+                enriched_pts_in_derivs = []
+                for elements_tups in zip_longest(*elements_deriv, fill_value=[]):
+                    # TODO: Combine repeated points?
+                    # TODO: Remove empty tuples?
+                    for tups in elements_tups:
+                        try:
+                            elements_point_set, weight_tensor, alpha_tensor = elements_tups
+                        except ValueError:  # Empty
+                            continue
+
+                    enriched_pts_in_derivs.append(chain(*elements_tups))
+                enriched_derivs.append(tuple(enriched_pts_in_derivs))
+            enriched_dual_basis.append(tuple(enriched_derivs))
+        return tuple(enriched_dual_basis)
+
     @property
     def mapping(self):
         mappings = set(elem.mapping for elem in self.elements)
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -130,7 +130,6 @@ class EnrichedElement(FiniteElementBase):
 
     @property
     def dual_basis(self):
-        # Note: NOT TESTED!!!
         elements_dual_basis = [element.dual_basis for element in self.elements]
         return tuple(chain(*elements_dual_basis))
 
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -1,4 +1,5 @@
 from functools import partial
+from itertools import chain
 from operator import add, methodcaller
 
 import numpy
@@ -129,8 +130,6 @@ class EnrichedElement(FiniteElementBase):
 
     @property
     def dual_basis(self):
-        from itertools import chain
-
         # Note: NOT TESTED!!!
         elements_dual_basis = [element.dual_basis for element in self.elements]
         return tuple(chain(*elements_dual_basis))
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -1,5 +1,4 @@
 from functools import partial
-from itertools import chain
 from operator import add, methodcaller
 
 import numpy
@@ -130,8 +129,7 @@ class EnrichedElement(FiniteElementBase):
 
     @property
     def dual_basis(self):
-        elements_dual_basis = [element.dual_basis for element in self.elements]
-        return tuple(chain(*elements_dual_basis))
+        return tuple(dual_vector for element in self.elements for dual_vector in element.dual_basis)
 
     @property
     def mapping(self):
--- a/finat/fiat_elements.py
+++ b/finat/fiat_elements.py
@@ -1,5 +1,6 @@
 import numpy as np
 import sympy as sp
+import collections
 from functools import singledispatch
 
 import FIAT
@@ -173,133 +174,139 @@ class FiatElement(FiniteElementBase):
         # Dispatch on FIAT element class
         return point_evaluation(self._element, order, refcoords, (entity_dim, entity_i))
 
-    def dual_evaluation(self, fn, entity=None):
-        '''Return code for performing the dual evaluation at the nodes of the
-        reference element. Currently only works for point evaluation and quadrature.
+    def dual_basis(self):
+        '''Returns a tuple where each element of the tuple represents one
+        functional in the dual shape. Each functional is represented by
+        a tuple of tuple containing the points (PointSet), weights
+        (Indexed Literal), multi_indices for extracting alpha from ReferenceGrad
+        (tuple), and components (tuple), sorted by total derivative order.
+        Can be thought of as a tuples of tuples of flattened tuples to avoid
+        messy dictionary key-item pairings.
 
-        :param fn: Function that takes in PointSet and returns GEM expression.
-        :param entity: the cell entity on which to tabulate for comparing
-                       results with FIAT.
+        For example, a dual basis containing 2 functionals with maximum derivative
+        order of 1 would be represented by:
+        (((point_set_10, weights_10, multi_indices_10, cmp_10),
+          (point_set_11, weights_11, multi_indices_11, cmp_11))
+         ((point_set_20, weights_20, multi_indices_20, cmp_20),
+          ()))
+        where one of the innermost tuples are empty because there are no evaluations
+        correponding to that order.
         '''
-        # TODO: Separate dual extraction _method from weight _method
-
-        import collections
         from itertools import chain
         from finat.point_set import PointSet
 
         max_deriv_order = max([ell.max_deriv_order for ell in self._element.dual_basis()])
-        # if max_deriv_order > 0:
-        #     raise NotImplementedError("Can only interpolate onto dual basis functionals without derivative evaluation, sorry!")
 
+        duals = []
+        point_set_cache = {}    # To avoid repeating points?
+        for dual in self._element.dual_basis():
+            derivs = []
+            # No of points = no of evaluations x no of (points as keys)
+            tup = tuple(sorted((pt, wt, cmp) for pt, tup in dual.get_point_dict().items()
+                               for (wt, cmp) in tup))
+            if len(tup) != 0:
+                pts, weights, cmp = zip(*tup)
+                alphas = tuple(tuple() for _ in pts)
+
+                try:
+                    point_set = point_set_cache[(pts)]
+                except KeyError:
+                    point_set = PointSet(pts)
+                    point_set_cache[(pts)] = point_set
+                weights = gem.Indexed(gem.Literal(weights), point_set.indices)
+                # TODO: Turn into GEM object?
+                # alphas = None
+                # cmp = None
+                derivs.append((point_set, weights, alphas, cmp))
+            else:
+                derivs.append(tuple())
+
+            for i in range(1, max_deriv_order+1):
+                tup = tuple(sorted((pt, alpha, wt, cmp) for pt, tup in dual.deriv_dict.items()
+                            for (wt, alpha, cmp) in tup if sum(alpha) == i))
+                if len(tup) != 0:
+                    pts, alphas, weights, cmp = zip(*tup)
+                    multi_indices = tuple(tuple(chain(*[(i,)*x for i, x in enumerate(alpha)]))
+                                          for alpha in alphas)
+
+                    try:
+                        point_set = point_set_cache[(pts)]
+                    except KeyError:
+                        point_set = PointSet(pts)
+                        point_set_cache[(pts)] = point_set
+                    weights = gem.Indexed(gem.Literal(weights), point_set.indices)
+                    # TODO: Turn into GEM object?
+                    # multi_indices = None
+                    # cmp = None
+
+                    derivs.append((point_set, weights, multi_indices, cmp))
+                else:
+                    derivs.append(tuple())
+            duals.append(tuple(derivs))
+        return tuple(duals)
+
+    def dual_evaluation(self, fn, entity=None):
+        '''Return code for performing the dual evaluation at the nodes of the
+        reference element. Currently only works for point evaluation and quadrature.
+
+        :param fn: Callable that takes in PointSet and returns GEM expression.
+        :param entity: the cell entity on which to tabulate for comparing
+                       results with FIAT.
+        '''
         if entity is None:
             # TODO: Add comparison to FIAT
             pass
 
-        # This is general code but is more unrolled than necessary.
-        # Creates expressions in order of derivative order and alpha, then combine
-        # together into one expression, then combine with weights and component
-        dual_expressions = []   # one for each functional
+        dual_expressions = []   # One for each functional
         expr_cache = {}         # Sharing of evaluation of the expression at points
-        for dual in self._element.dual_basis():
-            # Put each pt in corresponding derivative order
-            pt_dict = collections.OrderedDict()
-            pt_dict[0] = tuple(sorted((pt, ()) for pt in dual.get_point_dict().keys()))
-            # TODO: Can probably be optimized
-            # TODO: Fix case where one point has more than one alpha
-            for i in range(1, max_deriv_order+1):
-                pt_dict[i] = tuple(sorted((pt, alpha) for pt, tup in dual.deriv_dict.items()
-                                          for (wt, alpha, cmp) in tup if sum(alpha) == i))
-
-            # TODO: Remove repeated points
-            expr = gem.Zero()
-            for i in range(max_deriv_order+1):
-                print(i)
-                pts, alphas = zip(*pt_dict[i]) if len(pt_dict[i]) > 0 else (tuple(), tuple())
-                if len(pts) == 0:
-                    continue
-                try:
-                    expr, point_set = expr_cache[(pts, alphas)]
-                except KeyError:
-                    print('pts', pts)
-                    print('alphas', alphas)
-                    point_set = PointSet(pts)
-                    expr_grad = fn(point_set, derivative=i)
-                    print('expr_grad', expr_grad, expr_grad.shape)
-
-                    # Each point with derivative may have different alpha
-                    if i > 0:
-                        multi_indices = tuple(tuple(chain(*[(i,)*x for i, x in enumerate(alpha)])) for alpha in alphas)
-                        # How to separate alpha for each point?
-                        # Need to have same indices as weights
-                        # What if more than one point?
-                        # Flatten, is just sum of 'same' terms as i=0
-                        exprs_alpha = gem.ListTensor([gem.partial_indexed(expr_grad, idx) for idx in multi_indices])
-                        # exprs_alpha = gem.Indexed(exprs_alpha, (gem.Index() for _ in range(len(multi_indices))))
-
-                        # Combine for each point
-                        x = gem.Index()
-                        expr_alpha = gem.index_sum(gem.Indexed(exprs_alpha, (x, )), (x, ))
-                        print('exprs_alpha', exprs_alpha, exprs_alpha.shape)
-                        print('expr_alpha', expr_alpha, expr_alpha.shape)
-                        # What indices to sum over?
-                        expr = gem.Sum(expr, expr_alpha)
-                    else:
-                        expr = expr_grad
-
-                    # Hack to get shape_indices from shape of GEM expression
-                    # Unsure whether expressions with arguments work
-                    # Assuming general for all expressions including derivatives
-                    try:
-                        shape_indices
-                    except NameError:
-                        # Relies on first expression being order 0
-                        broadcast_shape = len(expr.shape) - len(self.value_shape)
-                        shape_indices = tuple(gem.Index() for _ in self.value_shape[:broadcast_shape])
-
-                    # TODO: Reshape for derivatives, must combine before reshape, but must match alpha indices with cmp
-                    print('expr', expr, expr.shape)
-                    print('broadcast_shape', broadcast_shape)
-                    print('shape_indices', shape_indices)
-                    expr = gem.partial_indexed(expr, shape_indices)
-                    expr_cache[(pts, alphas)] = expr, point_set
-
-        for dual in self._element.dual_basis():
-            # Put each pt in corresponding derivative order
-            pt_dict = collections.OrderedDict()
-            pt_dict[0] = tuple(sorted((pt, ()) for pt in dual.get_point_dict().keys()))
-            # TODO: Can probably be optimized
-            for i in range(1, max_deriv_order+1):
-                pt_dict[i] = tuple(sorted((pt, alpha) for pt, tup in dual.deriv_dict.items()
-                                          for (wt, alpha, cmp) in tup if sum(alpha) == i))
+        # Creates expressions in order of derivative order, extracts and sums alphas
+        # and components, then combine with weights
+        for dual in self.dual_basis():
             qexprs = gem.Zero()
-            for i in range(max_deriv_order+1):
-                pts, alphas = zip(*pt_dict[i]) if len(pt_dict[i]) > 0 else (tuple(), tuple())
-                if len(pts) == 0:
+            for i, deriv in enumerate(dual):
+                try:
+                    point_set, weights, multi_indices, cmp = deriv
+                except ValueError:  # Empty
                     continue
-                expr, point_set = expr_cache[(pts, alphas)]
 
-                weights = collections.defaultdict(list)
-                # TODO: replace by putting in all points and duals at once
-                if i == 0:
-                    for pt in pts:
-                        for (w, cmp) in dual.get_point_dict()[pt]:
-                            weights[(cmp, ())].append(w)
-                else:
-                    for pt in pts:
-                        for (w, alpha, cmp) in dual.deriv_dict[pt]:
-                            weights[(cmp, alpha)].append(w)
+                try:
+                    expr = expr_cache[(point_set, multi_indices)]
+                except KeyError:
+                    expr_grad = fn(point_set, derivative=i)
+                    # Extract derivative component
+                    # exprs_alpha = gem.ListTensor([gem.partial_indexed(expr_grad, idx) for idx in multi_indices])
 
-                # TODO: Fix order of alpha to order of cmp and pts???
-                for (cmp, alpha) in sorted(weights):
-                    print(cmp, alpha)
-                    print(weights[(cmp, alpha)])
-                    print(expr.shape)
-                    # TODO: case with more than one point/alpha
-                    qweights = gem.Literal(weights[(cmp, alpha)])
-                    qexpr = gem.Indexed(expr, cmp)
-                    qexpr = gem.index_sum(gem.Indexed(qweights, point_set.indices)*qexpr,
-                                          point_set.indices)
-                    qexprs = gem.Sum(qexprs, qexpr)
+                    # Extract derivative and component
+                    expr = gem.Zero()
+                    for idx, c in zip(multi_indices, cmp):
+                        # Insert shape indices between derivative and component?
+                        expr = gem.Sum(expr, gem.partial_indexed(gem.partial_indexed(expr_grad, idx), c))
+
+                    print(np.shape(multi_indices), multi_indices)
+                    print(np.shape(cmp), cmp)
+                    print(weights.shape, weights)
+                    # print(exprs_alpha.shape)
+                    # x = (gem.Index(), )
+                    # expr = gem.index_sum(gem.partial_indexed(exprs_alpha, x), x)
+
+                # Hack to get shape_indices from shape of GEM expression
+                # Unsure whether expressions with arguments work
+                # Assuming general for all expressions including derivatives
+                try:
+                    shape_indices
+                except NameError:
+                    broadcast_shape = len(expr.shape) - len(self.value_shape)
+                    shape_indices = tuple(gem.Index() for _ in self.value_shape[:broadcast_shape])
+
+                # Ignore arguments, move to between derivative and component?
+                expr = gem.partial_indexed(expr, shape_indices)
+                expr_cache[(point_set, multi_indices)] = expr
+
+                # Apply weights
+                qexpr = gem.index_sum(weights*expr, point_set.indices)
+                # Sum for all derivatives
+                qexprs = gem.Sum(qexprs, qexpr)
+
             assert qexprs.shape == ()
             dual_expressions.append(qexprs)
         ir_shape = gem.ListTensor(dual_expressions)
--- a/finat/fiat_elements.py
+++ b/finat/fiat_elements.py
@@ -198,12 +198,12 @@ class FiatElement(FiniteElementBase):
         duals = []
         point_set_cache = {}    # To avoid repeating points?
         for dual in self._element.dual_basis():
-            print(dual.deriv_dict)
+            # print(dual.get_point_dict(), dual.deriv_dict)
             derivs = []
             # No of points = no of evaluations x no of (points as keys)
             # tup = tuple(sorted((pt, wt, cmp) for pt, tup in dual.get_point_dict().items()
             #                    for (wt, cmp) in tup))
-            # TODO: repeated points
+            pts_in_derivs = []
             for pts, tups in sorted(dual.get_point_dict().items()):
                 try:
                     point_set = point_set_cache[(pts,)]
@@ -224,11 +224,13 @@ class FiatElement(FiniteElementBase):
                         weight_array[key] = item
                     weight_tensor = gem.Literal(weight_array)
 
-                derivs.append((point_set, weight_tensor, alpha_tensor))
+                pts_in_derivs.append((point_set, weight_tensor, alpha_tensor))
+            derivs.append(pts_in_derivs)
 
             deriv_dict_items = sorted(dual.deriv_dict.items())  # Ensure parallel safety
             for points, tups in deriv_dict_items:
                 weights, alphas, cmps = zip(*tups)
+                pts_in_derivs = []
                 for i in range(1, max_deriv_order+1):
                     # TODO: repeated points and repeated tups
                     weight, alpha, cmp = [], [], []
@@ -264,7 +266,8 @@ class FiatElement(FiniteElementBase):
                             weight_array[key] = item
                         weight_tensor = gem.Literal(weight_array)
 
-                    derivs.append((point_set, weight_tensor, alpha_tensor))
+                    pts_in_derivs.append((point_set, weight_tensor, alpha_tensor))
+                derivs.append(pts_in_derivs)
             duals.append(tuple(derivs))
         return tuple(duals)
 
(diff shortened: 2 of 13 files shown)


wence-:
So this presently migrates the hand-coded unrolling of FIAT dual basis data structures that was living in tsfc into FInAT. This is OK, but I think we should split the `dual_evaluation` code, which I think is fully generic if you have a dual basis object the right format, and the construction of the `dual_basis` expressions.

For duals without deriv_dict members, I think that all of the functionals are expressed as quadrature rules of some kind, so I would envisage that one has a `dual_basis` method that returns some gem expressions for the points and weights (presently these are unpicked in the dual_evaluation method).

WDYT?

dham:
And what does `dual_basis` return if there are derivatives?

wence-:
Points and weights to be applied to derivatives? I guess there's no way of distinguishing these cases. One could do it like fiat and return a dict whose keys are derivatives in directions (as tuples) and values are the points/weights expressions.

wence-:
I'm trying to cook up a way of separating the evaluation from the construction of the dual basis (such that tensor producting and the like compose well), but maybe I'm approaching it from the wrong direction.

rckirby:
First, this should all take into account the recent FIAT merge of the
`toriesz` branch that greatly expands the pt_dict and deriv_dict and keeps
functionals from having to override the to_riesz function and shell out to
sympy.

If deriv_dict (in FIAT) is empty, then this is no problem, and dual
evaluation can DTRT pretty easily on anything that supports point values.

If deriv_dict is not empty, then there is a problem.  As a first step, you
can raise an exception.  Other options: Is it possible to sympy-ize the
input, differentiate it, etc?  Otherwise, it may be messy (but possible) to
teach FIAT how to turn deriv_dict into more pt_dict business by means of
finite differences on the underlying space, but these seems hard.

mkan88:
@wence- What do you mean by being unable to distinguish the derivative cases?

I'm not quite sure how to deal with derivatives yet. Would the derivatives in UFL be of any use here?

dham:
I think I see how to do derivatives. The thing which gets passed into `dual_evaluation` is a function which takes a point set as an argument. It can also take an optional argument to indicate the derivative to be taken. In practice this can be evaluated in the callback using UFL differentiation, but that bit is not FInAT's problem and the calling routine could do differentiation in another way if it wanted.

mkan88:
Creating an explicit dual basis of GEM expressions appears to be quite tedious, if not hard. Are there any circumstances which the explicit GEM expressions rather than points and weights might be required?

I've added some sort of derivative functionality to the callback function in TSFC through UFL, but I'm not too sure how `dx` and `diff` would work, since they require a free index and `variable` respectively. GEM has equivalents to them, but passing them to the callback function should require a conversion back to UFL if I'm not mistaken.

dham:
Well we have a function of one vector variable. I think there are two sensible ways we could pass the request for a derivative. We either pass an integer and the function returns all derivatives of that degree as additional dimensions on the tensor, or we pass a multi-index (e.g. in 2d to get a first derivative in the x direction you pass (1,0)). I think it's open to debate which of these is better.

This is a much less general case than UFL differentiation so we don't have issues with indices etc.

wence-:
So I guess deciding whether or not to ask for directional derivatives or the whole thing depends partly how you'll use stuff later. I'm inclined to say that one requests the full jet up to a given order and then gets back an appropriately rank-augmented tensor. I think that inlining fixed indexing into this thing will DTRT more easily than trying to glue together x and y components of a derivative if one needs both.

miklos1:
> Creating an explicit dual basis of GEM expressions appears to be quite tedious, if not hard. Are there any circumstances which the explicit GEM expressions rather than points and weights might be required?

IIRC one of the (remote) goals of doing dual evaluation with FInAT is to enable possible optimisations, e.g., sum factorisation.  For example, interpolating a tensor product element to another tensor product element of different order should be sum factorisable.

It will probably take quite some work until everything works perfectly and efficiently, but the existing nodes `ListTensor` and `Concatenate` and their surrounding infrastructure may likely be of some help.

wence-:
>>  Creating an explicit dual basis of GEM expressions appears to be quite tedious, if not hard. Are there any circumstances which the explicit GEM expressions rather than points and weights might be required?

> IIRC one of the (remote) goals of doing dual evaluation with FInAT is to enable possible optimisations, e.g., sum factorisation. For example, interpolating a tensor product element to another tensor product element of different order should be sum factorisable.

Yes, consider interpolation into a Qk element on quads, the interpolation matrix is the kronecker product of two Qk interpolation matrices on intervals. In the current mode of implementation, this product would have to be "hand-coded" into a  `dual_evaluation` routine (plausibly recoding a bunch of the stuff that is in the `dual_evaluation` routine you now have).

dham:
> > > Creating an explicit dual basis of GEM expressions appears to be quite tedious, if not hard. Are there any circumstances which the explicit GEM expressions rather than points and weights might be required?
> 
> > IIRC one of the (remote) goals of doing dual evaluation with FInAT is to enable possible optimisations, e.g., sum factorisation. For example, interpolating a tensor product element to another tensor product element of different order should be sum factorisable.
> 
> Yes, consider interpolation into a Qk element on quads, the interpolation matrix is the kronecker product of two Qk interpolation matrices on intervals. In the current mode of implementation, this product would have to be "hand-coded" into a `dual_evaluation` routine (plausibly recoding a bunch of the stuff that is in the `dual_evaluation` routine you now have).

I think this line of commenting is just getting ahead of where we are. I agree that implementing tensor product and enriched and HDiv and HCurl is going to result in this code getting refactored. I'm less convinced that the right refactoring is into a dual_basis and a dual_evaluation.

mkan88:
Could I ask why a `dual_basis` method would not be a good idea?

finat/enriched.py
```diff
@@ -127,6 +128,11 @@ def point_evaluation(self, order, refcoords, entity=None):
                    for element in self.elements]
         return self._compose_evaluations(results)
 
+    @property
+    def dual_basis(self):
+        elements_dual_basis = [element.dual_basis for element in self.elements]
+        return tuple(chain(*elements_dual_basis))
```
wence-, line 134: style: inline the list comprehension.

finat/fiat_elements.py
```diff
@@ -142,6 +144,98 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                        if 0.0 not in weight:
+                            assert np.isclose(min(weight), max(weight))
+                        else:
+                            non_zero_weight = [w for w in weight if w != 0.0]
+                            assert np.isclose(min(non_zero_weight), max(non_zero_weight))
+
+                    # For direct indexing
+                    alpha_idx = tuple(tuple(chain(*[(j,)*a for j, a in enumerate(alph)])) for alph in alpha)
+                    # TODO: How to ensure correct combination (indexing) with weight_tensor?
+                    alpha_arr = np.zeros((self._element.ref_el.get_spatial_dimension(),) * derivative_order)
```
wence-, line 213: `self.cell.get_spatial_dimension()`

finat/fiat_elements.py
```diff
@@ -142,6 +144,98 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                        alpha_arr[idx] = 1
+                    alpha_tensor = gem.Literal(alpha_arr)
+
+                    try:
+                        point_set = point_set_cache[(pt,)]
+                    except KeyError:
+                        point_set = PointSet((pt,))
+                        point_set_cache[(pt,)] = point_set
+
+                    weight_dict = {c: w for w, c in zip(weight, cmp)}
```
wence-, line 224: style: `dict(zip(cmp, weight))`

finat/fiat_elements.py
```diff
@@ -142,6 +144,98 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                    # Each entry of tensor is weight of that component
+                    if len(self.value_shape) == 0:
+                        weight_tensor = gem.Literal(weight_dict[tuple()])
+                    else:
+                        weight_array = np.zeros(self.value_shape)
+                        for key, item in weight_dict.items():
+                            weight_array[key] = item
+                        weight_tensor = gem.Literal(weight_array)
```
wence-, line 232: Longwinded way of saying `gem.Literal(np.asarray(weight))`?

finat/finiteelementbase.py
```diff
@@ -144,6 +145,112 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+        For example, a dual basis containing 2 functionals with maximum derivative
+        order of 1 would be represented by:
+
+        .. highlight:: python
+        .. code-block:: python
+
+            ((((point_set_10, weight_tensor_10, alpha_tensor_10),
+               (point_set_11, weight_tensor_11, alpha_tensor_11)), None)
+             (((point_set_20, weight_tensor_20, alpha_tensor_20),
+              ()), None)
```
wence-, line 166: This is not really a datastructure.

finat/finiteelementbase.py
```diff
@@ -144,6 +145,112 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                        # TODO: Choose hash method
+                        expr = expr_cache[(point_set.points.data.tobytes(), alpha_tensor.array.tobytes())]
+                    except KeyError:
+                        expr_grad = fn(point_set, derivative=derivative_order)
+                        # TODO: Multiple alpha at once
+                        if derivative_order == 0:
+                            expr = expr_grad
+                        else:
+                            # Extract derivative component
+                            alpha_idx = tuple(gem.Index(extent=fn.dimension) for _ in range(derivative_order))
```
wence-, line 213: This requires that `fn` satisfies a particular interface which the docstring does not advertise.

finat/finiteelementbase.py
```diff
@@ -144,6 +145,112 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                        expr_cache[(point_set.points.data.tobytes(), alpha_tensor.array.tobytes())] = expr
+
+                    # Apply weights
+                    # For point_set with multiple points
+                    if tensorfe_idx is None:
+                        zeta = tuple(idx for _ in range(len(point_set.points)) for idx in self.get_value_indices())
+                        qexpr = gem.index_sum(gem.partial_indexed(expr, zeta) * weight_tensor[zeta], point_set.indices + zeta)
+                    else:
+                        try:
+                            base_rank
```
wence-, line 232: style: no.

finat/hdivcurl.py
```diff
@@ -202,3 +226,78 @@ def select_hcurl_transformer(element):
...
+    # Assume: something x interval
+    assert len(element.factors) == 2
+    assert element.factors[1].cell.get_shape() == LINE
+
+    # Globally consistent edge orientations of the reference
+    # quadrilateral: rightward horizontally, upward vertically.
+    # Their rotation by 90 degrees anticlockwise is interpreted as the
+    # positive direction for normal vectors.
+    ks = tuple(fe.formdegree for fe in element.factors)
+    if ks == (0, 1):
```
wence-, line 241: This function requires that the weight is a `gem.Literal`. I suspect that that might not be the case forever.

finat/hdivcurl.py
```diff
@@ -202,3 +226,78 @@ def select_hcurl_transformer(element):
...
+        # Apply scalar weight to scalar "base" expression, which is
+        # the leftward-pointing normal on the y-aligned edges.
+        return lambda w: gem.Literal([-w.value, 0])
+    elif ks == (1, 0):
+        # Apply scalar weight to scalar "base" expression, which is
+        # the scalar value the upward-pointing normal on the
+        # x-aligned edges.
+        return lambda w: gem.Literal([0, w.value])
```
wence-, line 249: Perhaps do something like 
```python
assert w.shape == ()
gem.ListTensor([0, w])
```
etc..

finat/tensorfiniteelement.py
```diff
@@ -120,6 +120,23 @@ def _tensorise(self, scalar_evaluation):
...
+
+        # Note: Does not assert idx of base_dual_basis is None
+        if not self._transpose:
+            tensor_dual_basis = tuple((base_dual, idx)
+                                      for base_dual, _ in base_dual_basis
+                                      for idx in numpy.ndindex(self.value_shape[base_rank:]))
+        else:
+            tensor_dual_basis = tuple((base_dual, idx)
+                                      for idx in numpy.ndindex(self.value_shape[base_rank:])
+                                      for base_dual, _ in base_dual_basis)
```
wence-, line 136: This explicitly writes out all the component indices, rather than having something with deltas in it. Should I expect that? It seems like this would produce a sequence of statements in the resulting kernel, rather than a loop.

finat/finiteelementbase.py
mkan88: I agree that it's quite ugly. Do you have anything in mind which could be iterated over and be easy to create other dual bases from?

finat/tensorfiniteelement.py
mkan88: Using free indices for both arguments to delta should be possible, but I think that would require a specialised dual evaluation method. Perhaps some sort of switch to choose between the generic and specialised dual evaluation method?

finat/finiteelementbase.py
wence-: So I think I don't understand what the bits are and how you destructure them. Each dual basis is made of, I think, a "quadrature rule" (point set + weight expression pair) and a multiindex indicating which derivatives that evaluation needs. Then there's something with components?

finat/tensorfiniteelement.py
wence-: OK, I think I see what might have happened. You've written a single fully-generic dual eval method which accepts any dual basis and that has a particular interface. So here you're having to satisfy that interface. Given that you need composition of the duals, you can't have different elements specifying a `dual_basis` property that satisfy different interfaces. Which means that  the generic thing has to work.

I think right now the components are picked out by having explicit numeric indices, and then you use those to index the UFL expression being evaluated literally.

Would it work to do something like this (sketch pseudo-code). The `component` part of the dual is a function that takes as many free indices as the ufl expression has shape axes.

then the evaluation of a component looks something like (pseudo-code)

```
shaped_gem_expr = compile_ufl(ufl_expr, ..., pointset=pointset)
indices = gem.indices(len(ufl_expr.ufl_shape))

gem_expr = Indexed(shaped_gem_expr, indices)
component = dual.component(*indices)

expr = index_sum(component*gem_expr, indices)
```

What does `component` look like. Well, if we have a vector-valued thing, and want to pick out the third component, then `indices` will be length-1, say we get an index `i`. Then if `dual.component(i)` returns `Delta(i, 2)` we end up with:

```
\sum_i \delta_{i, 2} expr_i = expr_2 (as required)
```

wence-: One has to think about how to compose these functions, but that might work.

ReubenHill:
@wence- 

I have run the below code to check the kernels for point evaluation nodes:

```python
from firedrake import *

m = UnitSquareMesh(5,5)
V = FunctionSpace(m, 'CG', 1)
W = FunctionSpace(m, 'CG', 2)
from functools import reduce
from operator import mul
expr = reduce(mul, SpatialCoordinate(m))
v = Function(V).project(expr)
w_v = interpolate(v, W)
```

On branch `firedrake` `tsfc` and `FInAT` branches `master` in `w_v = interpolate(v, W)` variable `kernel` at
`firedrake/interpolation.py(247)` is, accordign to pdb

```
Kernel("""---------------------------------------------------------------------------
KERNEL: expression_kernel
---------------------------------------------------------------------------
ARGUMENTS:
A: type: np:dtype('float64'), shape: (6), dim_tags: (N0:stride:1) aspace: global
w_0: type: np:dtype('float64'), shape: (3), dim_tags: (N0:stride:1) aspace: global
---------------------------------------------------------------------------
DOMAINS:
{ [p0] : 0 <= p0 <= 5 }
{ [i] : 0 <= i <= 2 }
---------------------------------------------------------------------------
INAME IMPLEMENTATION TAGS:
i: None
p0: None
---------------------------------------------------------------------------
TEMPORARIES:
t0: type: np:dtype('float64'), shape: (6, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
---------------------------------------------------------------------------
INSTRUCTIONS:
for i, p0
    A[p0] = A[p0] + t0[p0, i]*w_0[i]  {id=insn, priority=1}
end i, p0
---------------------------------------------------------------------------""", 'expression_kernel')
```

On branch `firedrake` `tsfc` and `FInAT` branches `dual_base_fiat` in `w_v = interpolate(v, W)` variable `kernel` at
`firedrake/interpolation.py(257)` is, according to pdb

```
Kernel("""---------------------------------------------------------------------------
KERNEL: expression_kernel
---------------------------------------------------------------------------
ARGUMENTS:
A: type: np:dtype('float64'), shape: (6), dim_tags: (N0:stride:1) aspace: global
w_0: type: np:dtype('float64'), shape: (3), dim_tags: (N0:stride:1) aspace: global
---------------------------------------------------------------------------
DOMAINS:
{ [i] : 0 <= i <= 2 }
{ [p0] : 0 <= p0 <= 5 }
---------------------------------------------------------------------------
INAME IMPLEMENTATION TAGS:
i: None
p0: None
---------------------------------------------------------------------------
TEMPORARIES:
t0: type: np:dtype('float64'), shape: (1, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
t1: type: np:dtype('float64'), shape: (1, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
t2: type: np:dtype('float64'), shape: (1, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
t3: type: np:dtype('float64'), shape: (1, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
t4: type: np:dtype('float64'), shape: (1, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
t5: type: np:dtype('float64'), shape: (1, 3), dim_tags: (N1:stride:3, N0:stride:1) scope:local
t6: type: np:dtype('float64'), shape: () scope:local
t7: type: np:dtype('float64'), shape: () scope:local
t8: type: np:dtype('float64'), shape: () scope:local
t9: type: np:dtype('float64'), shape: () scope:local
t10: type: np:dtype('float64'), shape: () scope:local
t11: type: np:dtype('float64'), shape: () scope:local
t12: type: np:dtype('float64'), shape: (6), dim_tags: (N0:stride:1) scope:local
---------------------------------------------------------------------------
INSTRUCTIONS:
↱  t11 = 0.0  {id=insn, priority=19}
└↱ t10 = 0.0  {id=insn_0, priority=18}
↱└ t9 = 0.0  {id=insn_1, priority=17}
└↱ t8 = 0.0  {id=insn_2, priority=16}
↱└ t7 = 0.0  {id=insn_3, priority=15}
└↱ t6 = 0.0  {id=insn_4, priority=14}
 │ for i
↱└   t6 = t6 + t5[0, i]*w_0[i]  {id=insn_5, priority=13}
└↱   t7 = t7 + t4[0, i]*w_0[i]  {id=insn_6, priority=12}
↱└   t8 = t8 + t3[0, i]*w_0[i]  {id=insn_7, priority=11}
└↱   t9 = t9 + t2[0, i]*w_0[i]  {id=insn_8, priority=10}
↱└   t10 = t10 + t1[0, i]*w_0[i]  {id=insn_9, priority=9}
└↱   t11 = t11 + t0[0, i]*w_0[i]  {id=insn_10, priority=8}
 │ end i
↱└ t12[0] = t11  {id=insn_11, priority=7}
└↱ t12[1] = t10  {id=insn_12, priority=6}
↱└ t12[2] = t9  {id=insn_13, priority=5}
└↱ t12[3] = t8  {id=insn_14, priority=4}
↱└ t12[4] = t7  {id=insn_15, priority=3}
└↱ t12[5] = t6  {id=insn_16, priority=2}
 │ for p0
 └   A[p0] = A[p0] + t12[p0]  {id=insn_17, priority=1}
   end p0
---------------------------------------------------------------------------""", 'expression_kernel')
```

I'm not sure exactly how to interpret kernel print statements but I see you mentioned above the code suggesting one would get a sequence of statements on this branch instead of a loop - it looks like this is what has happened, at least to some extent, here unless I am mistaken. I haven't gone through and carefully checked if the results are the same.

finat/hdivcurl.py
mkan88: `select_hdiv_transformer` and `select_hcurl_transformer` also require that the lambda input is a gem object.

ReubenHill:
@mkan88 are you planning to continue to actively develop these PRs or are you happy to hand over the push to merge to me?

finat/hdivcurl.py
wence-: Sure, but they don't make assumptions about what type of gem object it is.

wence-:
> I'm not sure exactly how to interpret kernel print statements but I see you mentioned above the code suggesting one would get a sequence of statements on this branch instead of a loop - it looks like this is what has happened, at least to some extent, here unless I am mistaken. I haven't gone through and carefully checked if the results are the same.

You can see what C code you would get out the other end with

```python
import loopy
print(loopy.generate_code_v2(kernel).device_code())
```

But yes, this looks like the new duals don't agglomerate things. If I do that for your examples I get:

Old
```c
void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };

  for (int32_t p0 = 0; p0 <= 5; ++p0)
    for (int32_t i = 0; i <= 2; ++i)
      A[p0] = A[p0] + t0[3 * p0 + i] * w_0[i];
}
```

New
```c
void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[3] = { 1.0, 0.0, 0.0 };
  double const t1[3] = { 0.0, 1.0, 0.0 };
  double t10;
  double t11;
  double t12[6];
  double const t2[3] = { 0.0, 0.0, 1.0 };
  double const t3[3] = { 0.0, 0.5, 0.5 };
  double const t4[3] = { 0.5, 0.0, 0.5 };
  double const t5[3] = { 0.5, 0.5, 0.0 };
  double t6;
  double t7;
  double t8;
  double t9;

  t11 = 0.0;
  t10 = 0.0;
  t9 = 0.0;
  t8 = 0.0;
  t7 = 0.0;
  t6 = 0.0;
  for (int32_t i = 0; i <= 2; ++i)
  {
    t6 = t6 + t5[i] * w_0[i];
    t7 = t7 + t4[i] * w_0[i];
    t8 = t8 + t3[i] * w_0[i];
    t9 = t9 + t2[i] * w_0[i];
    t10 = t10 + t1[i] * w_0[i];
    t11 = t11 + t0[i] * w_0[i];
  }
  t12[0] = t11;
  t12[1] = t10;
  t12[2] = t9;
  t12[3] = t8;
  t12[4] = t7;
  t12[5] = t6;
  for (int32_t p0 = 0; p0 <= 5; ++p0)
    A[p0] = A[p0] + t12[p0];
}
```

wence-:
Or if I make them `VectorFunctionSpace`:

Before

```c
void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };

  for (int32_t p0 = 0; p0 <= 5; ++p0)
    for (int32_t i = 0; i <= 1; ++i)
      for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
        A[2 * p0 + i] = A[2 * p0 + i] + t0[3 * p0 + i_0] * w_0[i + 2 * i_0];
}
```

After

```c
void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[3] = { 1.0, 0.0, 0.0 };
  double const t1[3] = { 0.0, 1.0, 0.0 };
  double t10[2];
  double t11[2];
  double t12[2];
  double t13[2];
  double t14[2];
  double t15[2];
  double t16[2];
  double t17[2];
  double const t18[2 * 2] = { 1.0, 0.0, 0.0, 1.0 };
  double t19;
  double const t2[3] = { 0.0, 0.0, 1.0 };
  double t20;
  double t21;
  double t22;
  double t23;
  double t24;
  double t25;
  double t26;
  double t27;
  double t28;
  double t29;
  double const t3[3] = { 0.0, 0.5, 0.5 };
  double t30;
  double t31[12];
  double const t4[3] = { 0.5, 0.0, 0.5 };
  double const t5[3] = { 0.5, 0.5, 0.0 };
  double t6[2];
  double t7[2];
  double t8[2];
  double t9[2];

  for (int32_t i = 0; i <= 1; ++i)
    t6[i] = 0.0;
  for (int32_t i_0 = 0; i_0 <= 1; ++i_0)
    t7[i_0] = 0.0;
  for (int32_t i_1 = 0; i_1 <= 1; ++i_1)
    t8[i_1] = 0.0;
  for (int32_t i_2 = 0; i_2 <= 1; ++i_2)
    t9[i_2] = 0.0;
  for (int32_t i_3 = 0; i_3 <= 1; ++i_3)
    t10[i_3] = 0.0;
  for (int32_t i_4 = 0; i_4 <= 1; ++i_4)
    t11[i_4] = 0.0;
  for (int32_t i_5 = 0; i_5 <= 1; ++i_5)
    t12[i_5] = 0.0;
  for (int32_t i_6 = 0; i_6 <= 1; ++i_6)
    t13[i_6] = 0.0;
  for (int32_t i_7 = 0; i_7 <= 1; ++i_7)
    t14[i_7] = 0.0;
  for (int32_t i_8 = 0; i_8 <= 1; ++i_8)
    t15[i_8] = 0.0;
  for (int32_t i_9 = 0; i_9 <= 1; ++i_9)
    t16[i_9] = 0.0;
  for (int32_t i_10 = 0; i_10 <= 1; ++i_10)
    t17[i_10] = 0.0;
  for (int32_t i_11 = 0; i_11 <= 2; ++i_11)
  {
    for (int32_t i_12 = 0; i_12 <= 1; ++i_12)
      t6[i_12] = t6[i_12] + t0[i_11] * w_0[2 * i_11 + i_12];
    for (int32_t i_13 = 0; i_13 <= 1; ++i_13)
      t7[i_13] = t7[i_13] + t0[i_11] * w_0[2 * i_11 + i_13];
    for (int32_t i_14 = 0; i_14 <= 1; ++i_14)
      t8[i_14] = t8[i_14] + t1[i_11] * w_0[2 * i_11 + i_14];
    for (int32_t i_15 = 0; i_15 <= 1; ++i_15)
      t9[i_15] = t9[i_15] + t1[i_11] * w_0[2 * i_11 + i_15];
    for (int32_t i_16 = 0; i_16 <= 1; ++i_16)
      t10[i_16] = t10[i_16] + t2[i_11] * w_0[2 * i_11 + i_16];
    for (int32_t i_17 = 0; i_17 <= 1; ++i_17)
      t11[i_17] = t11[i_17] + t2[i_11] * w_0[2 * i_11 + i_17];
    for (int32_t i_18 = 0; i_18 <= 1; ++i_18)
      t12[i_18] = t12[i_18] + t3[i_11] * w_0[2 * i_11 + i_18];
    for (int32_t i_19 = 0; i_19 <= 1; ++i_19)
      t13[i_19] = t13[i_19] + t3[i_11] * w_0[2 * i_11 + i_19];
    for (int32_t i_20 = 0; i_20 <= 1; ++i_20)
      t14[i_20] = t14[i_20] + t4[i_11] * w_0[2 * i_11 + i_20];
    for (int32_t i_21 = 0; i_21 <= 1; ++i_21)
      t15[i_21] = t15[i_21] + t4[i_11] * w_0[2 * i_11 + i_21];
    for (int32_t i_22 = 0; i_22 <= 1; ++i_22)
      t16[i_22] = t16[i_22] + t5[i_11] * w_0[2 * i_11 + i_22];
    for (int32_t i_23 = 0; i_23 <= 1; ++i_23)
      t17[i_23] = t17[i_23] + t5[i_11] * w_0[2 * i_11 + i_23];
  }
  t30 = 0.0;
  t29 = 0.0;
  t28 = 0.0;
  t27 = 0.0;
  t26 = 0.0;
  t25 = 0.0;
  t24 = 0.0;
  t23 = 0.0;
  t22 = 0.0;
  t21 = 0.0;
  t20 = 0.0;
  t19 = 0.0;
  for (int32_t i_24 = 0; i_24 <= 1; ++i_24)
    t19 = t19 + t6[i_24] * t18[2 * i_24];
  for (int32_t i_25 = 0; i_25 <= 1; ++i_25)
    t20 = t20 + t7[i_25] * t18[1 + 2 * i_25];
  for (int32_t i_26 = 0; i_26 <= 1; ++i_26)
    t21 = t21 + t8[i_26] * t18[2 * i_26];
  for (int32_t i_27 = 0; i_27 <= 1; ++i_27)
    t22 = t22 + t9[i_27] * t18[1 + 2 * i_27];
  for (int32_t i_28 = 0; i_28 <= 1; ++i_28)
    t23 = t23 + t10[i_28] * t18[2 * i_28];
  for (int32_t i_29 = 0; i_29 <= 1; ++i_29)
    t24 = t24 + t11[i_29] * t18[1 + 2 * i_29];
  for (int32_t i_30 = 0; i_30 <= 1; ++i_30)
    t25 = t25 + t12[i_30] * t18[2 * i_30];
  for (int32_t i_31 = 0; i_31 <= 1; ++i_31)
    t26 = t26 + t13[i_31] * t18[1 + 2 * i_31];
  for (int32_t i_32 = 0; i_32 <= 1; ++i_32)
    t27 = t27 + t14[i_32] * t18[2 * i_32];
  for (int32_t i_33 = 0; i_33 <= 1; ++i_33)
    t28 = t28 + t15[i_33] * t18[1 + 2 * i_33];
  for (int32_t i_34 = 0; i_34 <= 1; ++i_34)
    t29 = t29 + t16[i_34] * t18[2 * i_34];
  for (int32_t i_35 = 0; i_35 <= 1; ++i_35)
    t30 = t30 + t17[i_35] * t18[1 + 2 * i_35];
  t31[0] = t19;
  t31[1] = t20;
  t31[2] = t21;
  t31[3] = t22;
  t31[4] = t23;
  t31[5] = t24;
  t31[6] = t25;
  t31[7] = t26;
  t31[8] = t27;
  t31[9] = t28;
  t31[10] = t29;
  t31[11] = t30;
  for (int32_t p0 = 0; p0 <= 11; ++p0)
    A[p0] = A[p0] + t31[p0];
}
```

It might be that this is moot because the C compiler DTRT in both cases, but I would like to know.

ReubenHill:
> It might be that this is moot because the C compiler DTRT in both cases, but I would like to know.

You would like to know which: whether the C compiler optimises this out (which will likely depend on compiler optimisation options, whether there is a way to avoid this funny loop unrolling, or something else?

wence-:
> It might be that this is moot because the C compiler DTRT in both cases, but I would like to know.
You would like to know which: whether the C compiler optimises this out (which will likely depend on compiler optimisation options, whether there is a way to avoid this funny loop unrolling, or something else?

Whether the C compiler optimises this all away to start with.

I know there's a way to avoid the funny loop unrolling because the previous code avoids it.

Even if the C compiler manages to optimise things away, I would much prefer to have the unrolling not appear because it is destroying structure that we might want to use (e.g. these kernels are basically doing a matvec, we might want to use a library like libxsmm to make these calls at some point, but if the loops are gone in the gem representation already that is impossible).

mkan88:
> @mkan88 are you planning to continue to actively develop these PRs or are you happy to hand over the push to merge to me?

I'll be working on them part-time, but I'm happy to hand it over if you feel it would be more beneficial to do so.

finat/hdivcurl.py
mkan88: Would changing the input and output to be a simple tuple or list, and letting the dual basis method handle the conversion to a `Literal` be better?

finat/fiat_elements.py
mkan88: I think so; the dual basis for FIAT elements with derivatives is a bit muddy, it'll need some restructuring

finat/enriched.py
mkan88: Fixed

finat/fiat_elements.py
mkan88: Fixed

finat/finiteelementbase.py
mkan88: Added requirement for dimension attribute to docstring.

finat/finiteelementbase.py
mkan88: Fixed.

finat/tensorfiniteelement.py
mkan88: While we're thinking of how to do this for tensor finite elements, we should also try to do something like this for tensor product elements, since the current implementation does not exploit the product structure.

wence-: Is the way this new implementation works written up in your writeup? That might be a useful place to understand how to compose things correctly.

mkan88: Not quite, the report is mainly the mathematical representation, so implementation details are quite scarce.

wence-: I think this would still be helpful, effectively what we're trying to do here is to find an appropriate representation _in code_ for mathematical structure. It's hard to see this structure a priori in the code and deduce what it was in the maths.

ReubenHill:
> > It might be that this is moot because the C compiler DTRT in both cases, but I would like to know.
> > You would like to know which: whether the C compiler optimises this out (which will likely depend on compiler optimisation options, whether there is a way to avoid this funny loop unrolling, or something else?
> 
> Whether the C compiler optimises this all away to start with.
> 
> I know there's a way to avoid the funny loop unrolling because the previous code avoids it.
> 
> Even if the C compiler manages to optimise things away, I would much prefer to have the unrolling not appear because it is destroying structure that we might want to use (e.g. these kernels are basically doing a matvec, we might want to use a library like libxsmm to make these calls at some point, but if the loops are gone in the gem representation already that is impossible).

@sv2518 , @dham thinks you might know about some GEM parses that will get rid of unnecessary temporaries. Does anything spring to mind?

wence-:
> @sv2518 , @dham thinks you might know about some GEM parses that will get rid of unnecessary temporaries. Does anything spring to mind?

The problem is that many of the array-valued things are created as `ListTensor` objects (shaped things with no connection between the entries) which means that they have to be put in a temporary to be indexed. In a bunch of cases they should be `ComponentTensor` objects (shaped things that are a single expression that depends on indices).

finat/finiteelementbase.py
```diff
@@ -144,6 +145,110 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+              ()), None)
+
+        where one of the innermost tuples is empty because there are no evaluations
+        at that order. The `None` is either a placeholder for \'normal\' elements,
+        or a tuple containing the index of expression to extract using gem.Delta
+        (currently just used by TensorFiniteElement).
+        '''
+        # TODO: Add label for type of evaluation?
+
+    def dual_evaluation(self, fn, entity=None):
```
rckirby, line 175: Note: In order to get `PhysicallyMappedElement` (aka zany elements) correct, you need to overload `dual_evaluation`.  Morally, one should just get the list of results (functionals all applied to `fn`) and smack that with the transpose of the `PhysicallyMappedElement.basis_transformation` method.  To see how this is done with basis rather than dual evaluation: https://github.com/FInAT/FInAT/blob/2ef800c10efc37236a0e14f672caf9355b88a5d0/finat/physically_mapped.py#L158

ReubenHill:
So @dham and I came up with a new way of looking at the problem (please correct me if I'm regurgitating it wrong):

- All dual basis functionals  applied to some function  ultimately can be boiled down to weight tensors multiplied by a function point evaluations .
- The whole dual-basis-functional function evaluation problem can be represented as a system of equations. These often involve calculating the same function point evaluations  multiple times.
- We rewrite this system of equations as a (sparse) tensor of weight tensors  multiplied by the vector of 

1D Illustration (missing weights on the first and last functionals which are implicitly unity):
Screenshot 2020-12-15 at 16 17 36


New plan therefore:
1. We will capture the unstructured nature of the problem by looping over the quadrature rules that define each dual functional (loop over ) to build up a dictionary which maps points  to increasing point indices  as used in the function evaluations . At the same time we also build the vector of points .
2. Create from this dictionary a new (sparse) tensor of weight tensors ( above from ) in `finat.FiniteElementBase.dual_evaluation` - it will be a `gem.Literal`. Rows are basis functionals  and columns are point evaluation indices .
3. **LATER in a separate PR but before merging this** update `gem.Literal`s to be able to use the sparsity information that the dictionary provides, i.e. to deal with tensors which have sparse dimensions
4. We call the compiler on  to get 
5. We call `gem.IndexSum` on Q and  to contract out the index

wence-:
Sounds \phi-ne

ReubenHill:
@rckirby Hi Rob, David tells me that you’ve got work that shows that all dual functionals can really represented as weights multiplied by point evaluations, even where there are derivatives in functional definition. Is this correct and if so, can support for derivatives in this PR and https://github.com/firedrakeproject/tsfc/pull/218 be scrapped?

wence-:
> @rckirby Hi Rob, David tells me that you’ve got work that shows that all dual functionals can really represented as weights multiplied by point evaluations, even where there are derivatives in functional definition. Is this correct and if so, can support for derivatives in this PR and firedrakeproject/tsfc#218 be scrapped?

One would need to go round and replace all the functionals in FIAT that currently use derivatives with this trick. It effectively, I think, does some finite differencing so it might lose some precision and will also lose any structure that currently exists, I think.

ReubenHill:
> > I'm not sure exactly how to interpret kernel print statements but I see you mentioned above the code suggesting one would get a sequence of statements on this branch instead of a loop - it looks like this is what has happened, at least to some extent, here unless I am mistaken. I haven't gone through and carefully checked if the results are the same.
> 
> You can see what C code you would get out the other end with
> 
> ```python
> import loopy
> print(loopy.generate_code_v2(kernel).device_code())
> ```
> 
> But yes, this looks like the new duals don't agglomerate things. If I do that for your examples I get:
> 
> Old
> 
> ```c
> void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
> {
>   double const t0[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };
> 
>   for (int32_t p0 = 0; p0 <= 5; ++p0)
>     for (int32_t i = 0; i <= 2; ++i)
>       A[p0] = A[p0] + t0[3 * p0 + i] * w_0[i];
> }
> ```
> 
> New
> 
> ```c
> void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
> {
>   double const t0[3] = { 1.0, 0.0, 0.0 };
>   double const t1[3] = { 0.0, 1.0, 0.0 };
>   double t10;
>   double t11;
>   double t12[6];
>   double const t2[3] = { 0.0, 0.0, 1.0 };
>   double const t3[3] = { 0.0, 0.5, 0.5 };
>   double const t4[3] = { 0.5, 0.0, 0.5 };
>   double const t5[3] = { 0.5, 0.5, 0.0 };
>   double t6;
>   double t7;
>   double t8;
>   double t9;
> 
>   t11 = 0.0;
>   t10 = 0.0;
>   t9 = 0.0;
>   t8 = 0.0;
>   t7 = 0.0;
>   t6 = 0.0;
>   for (int32_t i = 0; i <= 2; ++i)
>   {
>     t6 = t6 + t5[i] * w_0[i];
>     t7 = t7 + t4[i] * w_0[i];
>     t8 = t8 + t3[i] * w_0[i];
>     t9 = t9 + t2[i] * w_0[i];
>     t10 = t10 + t1[i] * w_0[i];
>     t11 = t11 + t0[i] * w_0[i];
>   }
>   t12[0] = t11;
>   t12[1] = t10;
>   t12[2] = t9;
>   t12[3] = t8;
>   t12[4] = t7;
>   t12[5] = t6;
>   for (int32_t p0 = 0; p0 <= 5; ++p0)
>     A[p0] = A[p0] + t12[p0];
> }
> ```

For reference, with the present changes (making a big tensor of weights saved as a `gem.Literal` I get

```C
#include <stdint.h>

void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 6] = { 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0 };
  double const t1[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };
  double t2[6];

  for (int32_t i = 0; i <= 5; ++i)
  {
    t2[i] = 0.0;
    for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
      t2[i] = t2[i] + t1[3 * i + i_0] * w_0[i_0];
  }
  for (int32_t p0 = 0; p0 <= 5; ++p0)
    for (int32_t i_1 = 0; i_1 <= 5; ++i_1)
      A[p0] = A[p0] + t0[6 * p0 + i_1] * t2[i_1];
}
```

I'm not quite sure yet why I have two loops but I'm hoping that getting `gem.SparseLiteral` working properly will tidy this up.

ReubenHill:
I guess the extra loop is the `IndexSum` that does the tensor contraction

wence-:
> I guess the extra loop is the IndexSum that does the tensor contraction

It looks like `t0` is encoding the identity matrix. If it were `delta(i, i)` this might work?

wence-:
> It looks like `t0` is encoding the identity matrix. If it were `delta(i, i)` this might work?

Actually, why has `t0` appeared at all. This looks like what we did was generate `gem` to interpolate `w0` from P1 into P2 (`t1` is the interpolation matrix between P1 and P2), and then `gem` to generate that the interpolation of that expression into P2. But in such a way as to make it opaque to the optimiser.

ReubenHill:
It's because the weight tensor `Q` is the identity matrix here - I think I just need to notice that and replace it with delta as you suggest

ReubenHill:
Hmm, seems there's no shortcut that replaces the `gem.Identity` constant with `gem.Delta`

wence-:
> It's because the weight tensor Q is the identity matrix here - I think I just need to notice that and replace it with delta as you suggest.

Who made the `Q`? I guess it came from FIAT and was turned into a `gem.Identity`? Or is it coming from a call to `replace_delta`? Because the dual stuff makes some `Delta` nodes somewhere but no delta elimination is being done?

finat/finiteelementbase.py
```diff
@@ -144,6 +146,233 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+            expr_shape_indices = tuple(gem.Index(extent=ex) for ex in expr.shape)
+            assert Q.free_indices == ()
+            Q_shape_indices = tuple(gem.Index(extent=ex) for ex in Q.shape)
+            assert tuple(i.extent for i in Q_shape_indices[2:]) == tuple(i.extent for i in expr_shape_indices)
+            basis_indices = Q_shape_indices[:1]
+            dual_eval_is = gem.index_sum(Q[basis_indices + x.indices + expr_shape_indices] * expr[expr_shape_indices], x.indices+expr_shape_indices)
+            # Need to only have node count as free indices
+            dual_eval_is_w_shape = gem.ComponentTensor(dual_eval_is, basis_indices)
+            assert dual_eval_is_w_shape.shape[0] == Q.shape[0]
+            return dual_eval_is_w_shape
```
ReubenHill, line 316: > > It's because the weight tensor Q is the identity matrix here - I think I just need to notice that and replace it with delta as you suggest.
> 
> Who made the `Q`? I guess it came from FIAT and was turned into a `gem.Identity`? Or is it coming from a call to `replace_delta`? Because the dual stuff makes some `Delta` nodes somewhere but no delta elimination is being done?

I have made Q manually as outlined here


> So @dham and I came up with a new way of looking at the problem (please correct me if I'm regurgitating it wrong):
> 
> * All dual basis functionals  applied to some function ) ultimately can be boiled down to weight tensors multiplied by a function point evaluations ).
> * The whole dual-basis-functional function evaluation problem can be represented as a system of equations. These often involve calculating the same function point evaluations ) multiple times.
> * We rewrite this system of equations as a (sparse) tensor of weight tensors  multiplied by the vector of )
> 
> 1D Illustration (missing weights on the first and last functionals which are implicitly unity):
> Screenshot 2020-12-15 at 16 17 36
> 
> New plan therefore:
> 
> 1. We will capture the unstructured nature of the problem by looping over the quadrature rules that define each dual functional (loop over ) to build up a dictionary which maps points  to increasing point indices  as used in the function evaluations ). At the same time we also build the vector of points .
> 2. Create from this dictionary a new (sparse) tensor of weight tensors ( above from ) in `finat.FiniteElementBase.dual_evaluation` - it will be a `gem.Literal`. Rows are basis functionals  and columns are point evaluation indices .
> 3. **LATER in a separate PR but before merging this** update `gem.Literal`s to be able to use the sparsity information that the dictionary provides, i.e. to deal with tensors which have sparse dimensions
> 4. We call the compiler on  to get )
> 5. We call `gem.IndexSum` on  and ) to contract out the index 

This code shows my attempts to build it and tensor contract it with the point set.

ReubenHill: (edited for correctness)

wence-: OK. So the problem right now is that we're at the `Q` is a `gem.Literal` stage. I think that means things are probably fine. You could check by recognising this case and using `Q = gem.ComponentTensor(gem.Delta(i, j), (i, j))` I suppose. But if the later codegen step doesn't do delta elimination I guess we'll end up in the same place for now.

ReubenHill: IMO this should be refactored so that `dual_basis` returns a single weight tensor Q and point set x (to evaluate the expression at), rather than lots of point sets (each of which is always a single point(!)) and weight tensors in a huge tuple as present. 

This would avoid having to do a loop to build it each time (which takes ages for complex elements like BDM) and allow for case-by-case shortcuts like replacing an identity matrix with a delta.

ReubenHill: That is what I'm trying to do, the problem is that what's in code at the moment works out `Q` and a single point set at the `dual_evaluation` stage. I think this should be done at the `dual_basis` stage.

ReubenHill: Sadly that would mean redoing basically everything in this PR.

ReubenHill: > OK. So the problem right now is that we're at the `Q` is a `gem.Literal` stage. I think that means things are probably fine. You could check by recognising this case and using `Q = gem.ComponentTensor(gem.Delta(i, j), (i, j))` I suppose. But if the later codegen step doesn't do delta elimination I guess we'll end up in the same place for now.

If I insert this code

```python
            if is_identity:
                assert Q.shape[0] == Q.shape[1]
                i, j = gem.Index(extent=Q.shape[0]), gem.Index(extent=Q.shape[1])
                Q = gem.ComponentTensor(gem.Delta(i, j), (i, j))
```
the code generation isn't effected at all :/

finat/finiteelementbase.py
```diff
@@ -144,6 +146,233 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+
+            #
+            # TENSOR CONTRACT Q WITH expr
+            #
+            expr_shape_indices = tuple(gem.Index(extent=ex) for ex in expr.shape)
+            assert Q.free_indices == ()
+            Q_shape_indices = tuple(gem.Index(extent=ex) for ex in Q.shape)
+            assert tuple(i.extent for i in Q_shape_indices[2:]) == tuple(i.extent for i in expr_shape_indices)
+            basis_indices = Q_shape_indices[:1]
+            dual_eval_is = gem.index_sum(Q[basis_indices + x.indices + expr_shape_indices] * expr[expr_shape_indices], x.indices+expr_shape_indices)
```
miklos1, line 312: If you need to construct tensor contractions, please also consider using `gem.optimise.make_product` or `gem.optimise.sum_factorise`; they might give you extra benefits.

finat/finiteelementbase.py
wence-: I suspect very much that that is because the dual interpolation stuff in TSFC doesn't implement any of the structure-exploiting optimisations that assembly does. We basically construct the action of the dual and then feed it straight to scheduling.

ReubenHill:
```python
            if is_identity:
                assert Q.shape[0] == Q.shape[1]
                i, j = gem.Index(extent=Q.shape[0]), gem.Index(extent=Q.shape[1])
                j = i
                Q = gem.ComponentTensor(gem.Delta(i, j), (i, j))
```
does change the generated code
```C
#include <stdint.h>

void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };
  double t1;
  double t2;

  t2 = 0.0;
  for (int32_t i = 0; i <= 5; ++i)
  {
    t1 = 0.0;
    for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
      t1 = t1 + t0[3 * i + i_0] * w_0[i_0];
    t2 = t2 + t1;
  }
  for (int32_t p0 = 0; p0 <= 5; ++p0)
    A[p0] = A[p0] + t2;
}
```

but I get the wrong answer at the end since the indices are not correctly marked (I think). SO WIP!

wence-:
```
               assert Q.shape[0] == Q.shape[1]
                i, j = gem.Index(extent=Q.shape[0]), gem.Index(extent=Q.shape[1])
                j = i
                Q = gem.ComponentTensor(gem.Delta(i, j), (i, j))
```

This definitely doesn't do what you want it to. It makes a dense matrix of shape Q of all ones.

Your previous was correct, but the downstream consumer is not doing delta-elimination, so you end up getting the replacement of `delta_{i,j}` by an identity matrix. However, delivering `delta_{i,j}` here is _more right_ because that is the correct structure.

finat/finiteelementbase.py
ReubenHill: Replacing 
```python
dual_eval_is = gem.index_sum(Q[basis_indices + x.indices + expr_shape_indices] * expr[expr_shape_indices], x.indices+expr_shape_indices)
```
with
```python
dual_eval_is = gem.optimise.make_product((Q[basis_indices + x.indices + expr_shape_indices], expr[expr_shape_indices]), x.indices+expr_shape_indices)
```
change the produced code from
```C
#include <stdint.h>

void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 6] = { 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0 };
  double const t1[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };
  double t2[6];

  for (int32_t i = 0; i <= 5; ++i)
  {
    t2[i] = 0.0;
    for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
      t2[i] = t2[i] + t1[3 * i + i_0] * w_0[i_0];
  }
  for (int32_t p0 = 0; p0 <= 5; ++p0)
    for (int32_t i_1 = 0; i_1 <= 5; ++i_1)
      A[p0] = A[p0] + t0[6 * p0 + i_1] * t2[i_1];
}
```
to
```C
#include <stdint.h>

void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 6] = { 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0 };
  double const t1[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };
  double t2[6];

  for (int32_t i = 0; i <= 5; ++i)
  {
    t2[i] = 0.0;
    for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
      t2[i] = t2[i] + t1[3 * i + i_0] * w_0[i_0];
  }
  for (int32_t p0 = 0; p0 <= 5; ++p0)
    for (int32_t i_1 = 0; i_1 <= 5; ++i_1)
      A[p0] = A[p0] + t0[6 * p0 + i_1] * t2[i_1];
}
```
i.e. it makes no difference at all.

If I avoid wrapping the result back up with `dual_eval_is_w_shape = gem.ComponentTensor(dual_eval_is, basis_indices)` and instead output the `basis_indices` directly the generated code still doesn't change

```C
#include <stdint.h>

void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[6 * 6] = { 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0 };
  double const t1[6 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5, 0.5, 0.0, 0.5, 0.5, 0.5, 0.0 };
  double t2[6];

  for (int32_t i = 0; i <= 5; ++i)
  {
    t2[i] = 0.0;
    for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
      t2[i] = t2[i] + t1[3 * i + i_0] * w_0[i_0];
  }
  for (int32_t p0 = 0; p0 <= 5; ++p0)
    for (int32_t i_1 = 0; i_1 <= 5; ++i_1)
      A[p0] = A[p0] + t0[6 * p0 + i_1] * t2[i_1];
}
```

finat/finiteelementbase.py
ReubenHill: I don't follow? There's no dense matrix of ones in the generated code

wence-: Maybe I misinterpreted what was going on. But I think you want to get to a state where you're producing t2 with shape (6, ) not shape () (which looks like is what happened).

finat/finiteelementbase.py
wence-: I think you misinterpreted miklos' comment as being a "this will definitely fix things"? In this particular case that we're looking at now, the contraction is not complicated enough (or does not provide the structure such that) to be simplified by these routines.

However, if this is meant to be generic code, the comment is "if you're doing tensor contractions then you should consider delivering (potentially optimally scheduled) contractions by utilising the existing gem.optimise infrastructure" I think.

ReubenHill: Oooh it looks like there's delta cancellation in here! I'll try that

finat/finiteelementbase.py
ReubenHill: The simplest solution is probably to not do any tensor contraction at all when I have an identity matrix and just return `expr`

finat/finiteelementbase.py
ReubenHill: Aww it doesn't seem to kick in for some reason :/

rckirby:
You can take a look at the relevant FIAT bits in
`fiat/FIAT/pointwise_dual.py`.  This gets used in Hermite, Argyris, Morley
elements.  It's part of a larger deal that you need to come up with a
unisolvent set of points and then you can build the dual as linear
combinations of point values at those points.

Morally, it's finite differencing so will be exact on the function space
itself and inexact on an arbitrary callable.

miklos1:
IIRC `gem.Identity` was introduced to provide a code generation path for `gem.Delta`s when they are not eliminated.

> Aww it doesn't seem to kick in for some reason :/

Note that spectral optimisations (such as delta cancellation and sum factorisation) in TSFC do not work like chemotherapy, that they just kill all performance cancers throughout your expression body, but more like radiotherapy: because they are quite an aggressive method, they have to be pointed at the performance cancers that you need eliminating.  Namely, one needs to point at the (sub)expression to be optimised, determine how deep the factoriser may cut, and only then can the optimised expression be constructed.  These are currently configured for coefficient evaluation and top-level form assembly.

What I am saying is that you cannot just drop an optimisable expression at TSFC and expect it to magically become optimal.

wence-:
Some drive-by notes from the last firedrake meeting that Reuben wanted me to drop in.

Let us assume that the eventual interface ends up being that each finat element advertises a `dual_basis` property which returns two things:

1. An interpolation matrix `Q`, one row for each node in the finite element space, one column for each point in the point set
2. a `PointSet` (possibly with structure) which are the points at which the duals want expressions evaluated

(Let us argue later about what to do with derivative nodes).

Then some algebra on these things is available.

If I have a `TensorFiniteElement`, the `dual_basis` is obtained by (pseudo-code)
```python
Q, points = scalar.dual_basis
# Suppose the tensor element has shape (2, 4)
# These identity matrices may have difference sizes depending the shapes
# tQ = Q ⊗ 𝟙₂ ⊗ 𝟙₄
# This can be concretely realised with
indices = gem.indices(len(Q.shape))
i, j = (gem.Index(extent=2) for _ in range(2))
k, l = (gem.Index(extent=4) for _ in range(4))
# Need to check how this plays with the transpose argument to TensorFiniteElement.
tQ = gem.ComponentTensor(Q[*indices]*Delta(i, j)*Delta(k, l), indices + (i, j, k, l))
return tQ, points
```

Similarly a `TensorProductElement` outer-products the dual_basis of its factors
```
qs, pss = zip(*f.dual_basis for f in self.factors)
ps = TensorPointSet(pss)
# suppose there are two factors
qA, qB = qs
Aindices = gem.indices(len(qA.shape))
Bindices = gem.indices(len(qB.shape))
Q = gem.ComponentTensor(qA[*Aindices]*qB[*Bindices], Aindices + Bindices)
return Q, ps
```

Or something along those lines.

I suspect the moment nodes which each have some points that may have structure will complicate this. gem.Concatenate may be helpful here but I do not know.

ReubenHill:
@dham This may still need some fundamental design work. At the moment we have proposed a generic (sparse) tensor that encodes the weights of the entire dual basis which then acts upon a point set tensor and (roughly) sketched out how that would work in tensor product and tensor finite element cases.

@wence- points out that this may not capture all our use cases and we need to do a requirements capture. From slack:
> it's a really structured thing, because for each set of functionals that share evaluation points (e.g. all the internal moments for RT) you get a block in your sparse matrix (that may have internal structure)

>the eventual codegen for that will, I think, want to pick these blocks apart and do each one separately

>if you have a big sparse matrix then how do you communicate that structure?

ReubenHill:
Superseded by #89

PR #58, opened 2020-06-26
Direct serendipity

rckirby:
Adds a directly-defined serendipity quadrilateral element by Arbogast (https://arxiv.org/abs/1809.02192).  These include all polynomials of a given degree plus two rational functions.  They give optimal convergence rates on non-affine quads with minimal dimension.

This adds a new modality to FInAT -- "directly defined" elements that bypass a reference element and mapping and produce GEM for the basis functions directly in terms of a physical cell's vertices/coordinate system.  

This PR depends on PR #32 in FEniCS/ufl:  https://github.com/FEniCS/ufl/pull/32


finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+
+import gem
+import sympy
+from finat.sympy2gem import sympy2gem
+
+
+class DirectSerendipity(DirectlyDefinedElement, FiniteElementBase):
+    def __init__(self, cell, degree):
+        assert isinstance(cell, UFCQuadrilateral)
+        # assert degree == 1 or degree == 2
```
wence-, line 16: Remove comment.

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+from FIAT.polynomial_set import mis
+
+import gem
+import sympy
+from finat.sympy2gem import sympy2gem
+
+
+class DirectSerendipity(DirectlyDefinedElement, FiniteElementBase):
+    def __init__(self, cell, degree):
+        assert isinstance(cell, UFCQuadrilateral)
```
wence-, line 15: Can these be extended to higher dimensions, it's just you didn't do it? If so, do

```
if not isinstance(cell, UFCQuadrilateral):
    raise NotImplementedError(f"Not yet done for {cell} cells, please get in touch")
```

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+        else:
+            return {0: {i: [i] for i in range(4)},
+                    1: {i: list(range(4 + i * (self.degree-1),
+                                      4 + (i + 1) * (self.degree-1)))
+                        for i in range(4)},
+                    2: {0: list(range(4 + 4 * (self.degree - 1),
+                                      self.space_dim))}}
+
+    def space_dimension(self):
+        return self.space_dim
```
wence-, line 51: Just encode the formula here?

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+            return {0: {i: [i] for i in range(4)},
+                    1: {i: [i+4] for i in range(4)},
+                    2: {0: []}}
+        else:
+            return {0: {i: [i] for i in range(4)},
+                    1: {i: list(range(4 + i * (self.degree-1),
+                                      4 + (i + 1) * (self.degree-1)))
+                        for i in range(4)},
+                    2: {0: list(range(4 + 4 * (self.degree - 1),
+                                      self.space_dim))}}
```
wence-, line 48: Use `self.space_dimension()`

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+from finat.sympy2gem import sympy2gem
+
+
+class DirectSerendipity(DirectlyDefinedElement, FiniteElementBase):
+    def __init__(self, cell, degree):
+        assert isinstance(cell, UFCQuadrilateral)
+        # assert degree == 1 or degree == 2
+        self._cell = cell
+        self._degree = degree
+        self.space_dim = 4 if degree == 1 else (self.degree+1)*(self.degree+2)//2 + 2
```
wence-, line 19: No need to save this, just shove it in the `space_dimension` method (as commented below).

finat/physically_mapped.py
```diff
@@ -179,3 +192,21 @@ def physical_edge_lengths(self):
...
+
+    @abstractmethod
+    def physical_points(self, point_set, entity=None):
+        """Maps reference element points to GEM for the physical coordinates
+
+        :arg point_set: A point_set on the reference cell to push forward to physical space.
+        :arg entity: Reference cell entity on which the point set is
+             defined (for example if it is a point set on a facet).
+        :returns: a GEM expression for the physical locations of the
+        points, shape (gdim, ) with free indices of the point_set.
```
wence-, line 204: Indent this line by `     `

finat/sympy2gem.py
```diff
@@ -45,3 +45,8 @@ def sympy2gem_float(node, self):
 @sympy2gem.register(sympy.Symbol)
 def sympy2gem_symbol(node, self):
     return self.bindings[node]
+
+
+@sympy2gem.register(sympy.Rational)
+def sympy2gem_rational(node, self):
+    return gem.Literal(node.numerator()) / gem.Literal(node.denominator())
```
wence-, line 52: If they're both literals constant-fold them: `return gem.Literal(node.numerator() / node.denominator())`

wence-: But this seems oddly specific, it won't always be the case. So probably do
```
return gem.Division(self(node.numerator()), self(node.denominator()))
```

the `Division` node in gem already constant-folds literals.

wence-:
Use the citations-registration machinery to add a citation to the Arbogast paper/preprint. (see `physically_mapped.py` for how it's done (and then, for example, morley.py for how to use it).

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def mapping(self):
+        return "physical"
+
+
+def xysub(x, y):
+    return {x[0]: y[0], x[1]: y[1]}
+
+
+def ds1_sympy(ct):
```
wence-, line 111: Maybe some brief comments describing which equations in the paper you're implementing?

wence-: Also docstring, especially describing both the input argument, and the return value.

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+
+    phis = []
+    for i, phitilde in enumerate([phi0tilde, phi1tilde, phi2tilde, phi3tilde]):
+        phi = phitilde / phitilde.subs({xx[0]: vs[i, 0], xx[1]: vs[i, 1]})
+        phis.append(phi)
+
+    return vs, xx, numpy.asarray(phis)
+
+
+def newton_dd(nds, fs):
```
wence-, line 177: docstring.

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+    mat = numpy.zeros((n, n), dtype=object)
+    mat[:, 0] = fs[:]
+    for j in range(1, n):
+        for i in range(n-j):
+            mat[i, j] = (mat[i+1, j-1] - mat[i, j-1]) / (nds[i+j] - nds[i])
+    return mat[0, :]
+
+
+# Horner evaluation of polynomial in symbolic form
+def newton_poly(nds, fs, xsym):
```
wence-, line 188: Move to docstring, and describe arguments.

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+    coeffs = newton_dd(nds, fs)
+    result = coeffs[-1]
+    n = len(coeffs)
+    for i in range(n-2, -1, -1):
+        result = result * (xsym - nds[i]) + coeffs[i]
+    return result
+
+
+def dsr_sympy(ct, r, vs=None):
```
wence-, line 197: docstring.

finat/direct_serendipity.py
```diff
@@ -0,0 +1,371 @@
...
+
+                bfcur = (nodalize(prebf)
+                         / prebf.subs(xysub(xx, edge_nodes[ed][i])))
+                edge_bfs_cur.append(bfcur)
+
+            edge_bfs.append(edge_bfs_cur)
+
+    # vertex basis functions
+    vertex_to_adj_edges = {0: (0, 2), 1: (0, 3), 2: (1, 2), 3: (1, 3)}
+    vertex_to_off_edges = {0: (1, 3), 1: (1, 2), 2: (0, 3), 3: (0, 2)}
```
wence-, line 341: This encodes information about the fiat cell topology I think, can you deduce it from that, rather than hardcoding here (just in case we ever change anything).

finat/physically_mapped.py
wence-: Does this mean we need to augment `DirectlyDefined` elements with a `geometric_dimension` property? At the moment you can't tell, if you just have a finat element, what the embedding dimension is.

rckirby:
Not known.  Comment to that effect will be added.

rckirby:
Yes, got it and similar occurrences.

rckirby:
Perhaps so.  FInAT normally reads that out of FIAT (and DTRT for tensor
product, etc).  How do we do this?

rckirby:
Just return `self.cell.get_dimension())`?
But the DirectlyDefined class doesn't actually have an underlying cell.  It
could be added?

wence-:
The geometric dimension thing is kind of weird. You cant get it from the fiat cell because that doesn't know about any embedding space. But I guess we don't need to do anything because we can just ask the points for their shape

rckirby:
Ok.  Is any action required here then?  I think I took a stab at everything
else you mentioned.

finat/direct_serendipity.py
```diff
@@ -0,0 +1,405 @@
...
+    for i, phitilde in enumerate([phi0tilde, phi1tilde, phi2tilde, phi3tilde]):
+        phi = phitilde / phitilde.subs({xx[0]: vs[i, 0], xx[1]: vs[i, 1]})
+        phis.append(phi)
+
+    return vs, xx, numpy.asarray(phis)
+
+
+def newton_dd(nds, fs):
+    """Constructs Newt's divided differences for the input arrays, which may include
+    symoblic values."""
```
miklos1, line 189: - `Newt's`?
- `symoblic`

finat/physically_mapped.py
```diff
@@ -125,6 +139,12 @@ def point_evaluation(self, order, refcoords, entity=None):
         raise NotImplementedError("TODO: not yet thought about it")
 
 
+class DirectlyDefinedElement(NeedsCoordinateMappingElement):
+    """Base class for directly defined elements such as direct
+    serendipity that bypass a coordinate mapping."""
+    pass
```
miklos1, line 145: So this is just a marker class.  (It defines no useful interface, merely marks that the derived class is of a certain type.)  Fine.

finat/physically_mapped.py
```diff
@@ -179,3 +200,21 @@ def physical_edge_lengths(self):
...
+        :returns: a GEM expression for the physical locations of the
+                  points, shape (gdim, ) with free indices of the point_set.
+        """
+
+    @abstractmethod
+    def physical_vertices(self):
+        """Physical locations of the cell vertices.
+
+        :returns: a GEM expression for the physical vertices, shape
+                (gdim, )."""
```
miklos1, line 220: This could be implemented in terms of `physical_points`.  However, this non-minimality does not look too bad.

miklos1:
@rckirby: BTW do you have a test somewhere?  It is perhaps easiest to test it from Firedrake, but I think there should be some test somewhere which prevents this element from breaking silently in the future.

rckirby:
This is actually a tough question: How to test FInAT implementations when
they only generate GEM and without relying on tsfc + Firedrake?

I have done some convergence studies using Firedrake.  Maybe we should get
the various branches of FInAT/tsfc/ufl to land and then I can do a
Firedrake PR to incorporate tests?

miklos1:
> This is actually a tough question: How to test FInAT implementations when they only generate GEM and without relying on tsfc + Firedrake?

The standard technique is to provide stub implementations for the dependencies/callbacks, and then verify the immediate result. So in this case, you would write a `coordinate_mapping` implementation for the test case, that merely returns the correct GEM snippets that are expected for the sake of this specific test.  To verify the GEM expression that results from the basis evaluation, your best bet is probably to use the GEM interpreter and then just numerically compare the result with the known expected values.

> Maybe we should get the various branches of FInAT/tsfc/ufl to land and then I can do a Firedrake PR to incorporate tests?

That would work, yes. The above described method, i.e., testing just FInAT without testing the entire Firedrake stack, is better from a software engineering point of view, because then your tests are more fine-grained.  However, that is also the less trodden path thus far, so expect a few grievances and improvement opportunities on the way.

wence-:
> The standard technique is to provide stub implementations for the dependencies/callbacks, and then verify the immediate result. So in this case, you would write a coordinate_mapping implementation for the test case, that merely returns the correct GEM snippets that are expected for the sake of this specific test. To verify the GEM expression that results from the basis evaluation, your best bet is probably to use the GEM interpreter and then just numerically compare the result with the known expected values.

I think this is probably the preferred approach (we could then verify some of the other physicallymapped elements here too). @rckirby I can help with this next week, but you could give it a go assuming that you just want to do it on a reference element (so providing the relevant gem is easy)

rckirby:
Thanks, Lawrence.  Let's chat next week.

I think the right test is probably to just provide (GEM for) some physical
nodes and vertices and check Kronecker properties in FInAT.

rckirby:
OK, I got the test working on my machine, and it works off of the reference
element (I hand-coded the mapping from the reference cell to physical
space, maybe there is a better way?)

Somehow travis is unhappy with my code.  I will try to fix that and then we
can see about merging (we will need the tsfc bit to land as well as well as
ufl).

rckirby:
OK, it failing with "Object arrays are not supported" under Python 3.5 on
travis (works fine on 3.7.7 on my mac).
It wasn't failing before because there wasn't a test to execute that code.

Advice?

test/test_direct_serendipity.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def physical_vertices(self):
+        pvs = np.zeros(self.verts.shape, dtype=object)
+        for alpha in np.ndindex(pvs.shape):
+            pvs[alpha] = gem.Literal(self.verts[alpha])
+        return gem.ListTensor(pvs)
+
+
+def get_pts(cell, deg):
+    assert cell.shape == 11  # quadrilateral
```
miklos1, line 55: Please use the symbolic constant `QUADRILATERAL` instead of this magic number + comment.

test/test_direct_serendipity.py
```diff
@@ -0,0 +1,91 @@
...
+import finat
+import gem
+import numpy as np
+from finat.physically_mapped import PhysicalGeometry
+
+
+class MyMapping(PhysicalGeometry):
+    def __init__(self, cell, verts):
+        # cell is reference cell, verts is physical vertices
+        self.verts = np.asarray(verts)
```
miklos1, line 12: `self.cell` never assigned?

miklos1:
> OK, it failing with "Object arrays are not supported" under Python 3.5 on travis (works fine on 3.7.7 on my mac).

Perhaps a `.dot(` would do the job?

rckirby:
Wasn't used at all.  But I've assigned it in case any later revisions need
it.

rckirby:
Done

rckirby:
This can be done, but Lawrence also suggested bumping up to Python 3.6 will
make Travis happy.  Maybe dot is better (if uglier) in the off-hand chance
we have 3.5-based users?

miklos1:
> OK, it failing with "Object arrays are not supported" under Python 3.5 on travis (works fine on 3.7.7 on my mac).

Actually, this works for me with Python 3.5.  Perhaps outdated numpy version on Travis for Python 3.5?

miklos1:
> OK, it failing with "Object arrays are not supported" under Python 3.5 on travis (works fine on 3.7.7 on my mac).

Yes, I can reproduce the failure with Python 3.5.9 and numpy 1.15.4.  It is works with numpy 1.16.  I suggest dropping the fixing attempting (i.e. back to `@` operator), and simply asking for `numpy>=1.16` in requirements.

rckirby:
I bumped up the numpy version to 1.16 in the requirements, and travis is
passing with Python 3.5
Good catch, Miklos!

test/test_direct_serendipity.py
```diff
@@ -0,0 +1,81 @@
...
+    return pts
+
+
+@pytest.mark.parametrize('degree', [1, 2, 3, 4])
+def test_kronecker(degree):
+    cell = FIAT.ufc_cell("quadrilateral")
+    element = finat.DirectSerendipity(cell, degree)
+    pts = finat.point_set.PointSet(get_pts(cell, degree))
+    vrts = np.asarray(((0.0, 0.0), (1.0, 0.0), (0.1, 1.1), (0.95, 1.01)))
+    mppng = MyMapping(cell, vrts)
```
wence-, line 76: A definite fortran parentage in the naming conventions here.

miklos1: Renamed.

miklos1:
> Something has gone wrong with the elements for degree 4+ in the tests. Let's wait to merge till the Firedrake tests are passing (which means FInAT is fixed)

@rckirby: So what is the state of this?  Has FInAT since been changed to fix this?  I just force-pushed the firedrake branch for testing, and merged master into the all the branches to avoid errors due to branch inconsistencies...

@wence-: I renamed the Fortran variable names.  Are there any more objections to merging this?

miklos1:
> Something has gone wrong with the elements for degree 4+ in the tests.

I tried it locally and it is still failing...

rckirby:
I tracked and fixed the degree 4+ bug in FInAT.  We need to see if the
Firedrake tests can be made to pass

rckirby:
Latest commit (moving to symengine and changing the method of constructing internal Lagrange functions) seems to fix the long compile times reported for high degree and correctness in the related firedrake test suite.  Ready for last review and hopefully to merge soon.

PR #60, opened 2020-08-10
Implement RestrictedElement

wence-:
This pushes everything through the element algebra, rather than masking out dofs that we want to drop. I think this is the best way.

### DIFF
--- a/finat/restricted.py
+++ b/finat/restricted.py
@@ -1,0 +1,260 @@
+from functools import singledispatch
+from itertools import chain
+
+import FIAT
+from FIAT.polynomial_set import mis
+
+import finat
+from finat.fiat_elements import FiatElement
+from finat.physically_mapped import PhysicallyMappedElement
+
+
+# Sentinel for when restricted element is empty
+null_element = object()
+
+
+@singledispatch
+def restrict(element, domain, take_closure):
+    """Restrict an element to a given subentity.
+
+    :arg element: The element to restrict.
+    :arg domain: The subentity to restrict to.
+    :arg take_closure: Gather dofs in closure of the subentities?
+        Ignored for "interior" domain.
+
+    :raises NotImplementedError: If we don't know how to restrict this
+        element.
+    :raises ValueError: If the restricted element is empty.
+    :returns: A new finat element."""
+    return NotImplementedError(f"Don't know how to restrict element of type {type(element)}")
+
+
+@restrict.register(FiatElement)
+def restrict_fiat(element, domain, take_closure):
+    try:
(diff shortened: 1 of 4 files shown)


finat/restricted.py
```diff
@@ -0,0 +1,219 @@
...
+    elif d == 1:
+        return "edge"
+    elif d == 2:
+        return "face"
+    else:
+        raise ValueError
+
+
+@restrict.register(finat.TensorProductElement)
+def restrict_tpe(element, domain, take_closure):
```
pefarrell, line 152: It would be nice to add a comment here explaining the algorithm.

finat/restricted.py
```diff
@@ -0,0 +1,219 @@
...
+        if isinstance(restricted, finat.EnrichedElement):
+            return finat.EnrichedElement(finat.HDivElement(e)
+                                         for e in restricted.elements)
+        else:
+            return finat.HDivElement(restricted)
+
+
+@restrict.register(finat.mixed.MixedSubElement)
+def restrict_mixed(element, domain, take_closure):
+    raise NotImplementedError("Don't know how to handle mixed + restricted")
```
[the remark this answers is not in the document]
wence-: I think so, I have to handle it in `EnrichedElement` due to the way finat implements mixedelement.

finat/restricted.py
wence-: Done, hopefully makes sense.

wence-:
Suppose I should add a few tests.

finat/restricted.py
pefarrell: Yep, looks great.

pefarrell:
Agreed. Once there are some tests this is good to merge. Thanks!

finat/restricted.py
```diff
@@ -0,0 +1,260 @@
...
+
+    .. warning::
+
+       The returned element *may not* be interpolatory. That is, the
+       dual basis (if implemented) might not be nodal to the primal
+       basis. Assembly still works (``basis_evaluation`` is fine), but
+       interpolation may produce bad results.
+
+       Restrictions of FIAT-implemented CiarletElements are always
+       nodal.
```
wence-, line 252: Added some extra detail here warning about possible non-nodality.

wence-:
Added tests, happy?

pefarrell:
Looks great, thanks.

miklos1:
Related: this is one of two non-leaf elements (element algebra nodes) which make the FInAT interface fall back to the FIAT interface and just wrap the result.  (The other one is the nodal enriched element.)

Beyond preserving structure for restricted elements, this is also a helpful step towards retiring the FIAT interface.

PR #61, opened 2020-08-13
wence/docs

wence-:
Squash warnings in doc build, migrate testing to github actions and autopublish to gh-pages on successful master build.


wence-:
@rckirby I realised we were hosting finat docs, but they were very out of date. This branch fixes that (and makes sure they won't be in the future!)

PR #63, opened 2020-08-18
Some zany H(div) elements

rckirby:
FInAT level support for Mardal-Tai-Winther and Arnold-Winther (conforming and non-conforming) elements.  Relies on associated UFL (https://github.com/FEniCS/ufl/pull/38 and FIAT) and FIAT(https://github.com/FEniCS/fiat/pull/56) PR's to land first.  Associated TSFC PR coming soon will want this and UFL/FIAT PRs to land first.


miklos1:
- Normally, we don't want to have commented out code snippets in production code.  Leaving them around during development might be useful, but remove them when it comes to a PR.
- There are flake8 failures.

test/test_aw.py
```diff
@@ -0,0 +1,96 @@
...
+        for k in range(ref_vals.shape[3]):
+            ref_vals_piola[i, :, :, k] = \
+                J @ ref_vals[i, :, :, k] @ J.T / detJ**2
+
+    # Zany map the results
+    mppng = MyMapping(ref_cell, phys_cell)
+    Mgem = ref_el_finat.basis_transformation(mppng)
+    M = evaluate([Mgem])[0].arr
+    # print(M)
+    # print(ref_vals_piola.shape)
```
miklos1, line 38: Commented out code.

test/test_aw.py
```diff
@@ -0,0 +1,96 @@
...
+
+    ref_vals_piola = np.zeros(ref_vals.shape)
+    for i in range(ref_vals.shape[0]):
+        for k in range(ref_vals.shape[3]):
+            ref_vals_piola[i, :, :, k] = \
+                J @ ref_vals[i, :, :, k] @ J.T / detJ**2
+
+    # Zany map the results
+    mppng = MyMapping(ref_cell, phys_cell)
+    Mgem = ref_el_finat.basis_transformation(mppng)
```
miklos1, line 35: Probably `mapping` is preferable to `mppng`.

test/test_aw.py
```diff
@@ -0,0 +1,96 @@
...
+    M = evaluate([Mgem])[0].arr
+    # print(M)
+    # print(ref_vals_piola.shape)
```
miklos1, line 79: Commented out code.

test/test_aw.py
```diff
@@ -0,0 +1,96 @@
...
+
+    ref_vals_piola = np.zeros(ref_vals.shape)
+    for i in range(ref_vals.shape[0]):
+        for k in range(ref_vals.shape[3]):
+            ref_vals_piola[i, :, :, k] = \
+                J @ ref_vals[i, :, :, k] @ J.T / detJ**2
+
+    # Zany map the results
+    mppng = MyMapping(ref_cell, phys_cell)
+    Mgem = ref_el_finat.basis_transformation(mppng)
```
miklos1, line 76: `mppng`

test/test_aw.py
```diff
@@ -0,0 +1,96 @@
...
+                    M @ ref_vals_piola[:, ell1, ell2, k]
+
+    # Q = FIAT.make_quadrature(phys_cell, 6)
+    # vals = phys_element.tabulate(0, Q.pts)[(0, 0)]
+    # print()
+    # for i in (0, 9, 21):
+    #     result = 0.0
+    #     for k in range(len(Q.wts)):
+    #         result += Q.wts[k] * (vals[i, :, :, k] ** 2).sum()
+    #     print(np.sqrt(result))
```
miklos1, line 94: Commented out code.

test/test_morley.py
```diff
@@ -0,0 +1,25 @@
...
+    ref_cell = FIAT.ufc_simplex(2)
+    ref_element = finat.Morley(ref_cell, 2)
+    ref_pts = finat.point_set.PointSet(ref_cell.make_points(2, 0, 4))
+
+    phys_cell = FIAT.ufc_simplex(2)
+    phys_cell.vertices = ((0.0, 0.1), (1.17, -0.09), (0.15, 1.84))
+
+    mppng = MyMapping(ref_cell, phys_cell)
+    z = (0, 0)
+    finat_vals_gem = ref_element.basis_evaluation(0, ref_pts, coordinate_mapping=mppng)[z]
```
miklos1, line 18: `mppng`

test/test_morley.py
```diff
@@ -0,0 +1,25 @@
+import FIAT
+import finat
+import numpy as np
+from gem.interpreter import evaluate
+from fiat_mapping import MyMapping
+
+
+def test_morley():
```
miklos1, line 8: Morley is unrelated to this pull request, nevertheless, better test coverage is welcome.

miklos1:
Pushed a commit for CI testing, which had been failing due to not using the correct FIAT branch.

finat/physically_mapped.py
```diff
@@ -142,6 +181,14 @@ def jacobian_at(self, point):
...
+    def detJ_at(self, point):
+        """The determinant of the jacobian of the physical coordinates at a point.
+
+        :arg point: The point in reference space to evaluate the Jacobian determinant.
+        :returns: A GEM expression for the Jacobian determinant.
+        """
+
     @abstractmethod
     def reference_normals(self):
         """The (unit) reference cell normals for each facet.
```
miklos1, line 194: `reference_edge_tangents` is added in firedrakeproject/tsfc#225, but not here.

miklos1 (changes requested):

finat/aw.py
```diff
@@ -0,0 +1,158 @@
...
+        J_np = numpy.array([[J[0, 0], J[0, 1]],
+                            [J[1, 0], J[1, 1]]])
+        JTJ = J_np.T @ J_np
+
+        for e in range(3):
+
+            # Compute alpha and beta for the edge.
+            Ghat_T = numpy.array([nhat[e, :], that[e, :]])
+
+            (alpha, beta) = Ghat_T @ JTJ @ that[e, :] / detJ
```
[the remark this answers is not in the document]
wence-: firedrakeproject/tsfc#229 adds `@` magic method and `.T` property. So that I think this syntax would be possible without numpy.

test/test_aw.py
rckirby: fixed now.

test/test_aw.py
rckirby: changed.

test/test_aw.py
rckirby: removed.

test/test_aw.py
rckirby: removed

test/test_morley.py
rckirby: Yes, I was doing it as a sanity check for using the gem interpreter in tests.  Unrelated, but let's keep it in.

test/test_aw.py
rckirby: fixed.

test/test_morley.py
rckirby: fixed.

finat/aw.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # This bypasses the GEM wrapper.
+        that = numpy.array([T.compute_normalized_edge_tangent(i) for i in range(3)])
+        nhat = numpy.array([T.compute_normal(i) for i in range(3)])
+
+        detJ = coordinate_mapping.detJ_at([1/3, 1/3])
+        J = coordinate_mapping.jacobian_at([1/3, 1/3])
+        J_np = numpy.array([[J[0, 0], J[0, 1]],
+                            [J[1, 0], J[1, 1]]])
+        JTJ = J_np.T @ J_np
```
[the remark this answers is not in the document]
rckirby: The `V` matrix I'm building is a `numpy` array (this is also done in Hermite, Morley, etc) that gets converted to a `gem.ListTensor` at the end of the routine.  Is there a way to avoid this?  If I can keep everything at the `gem` level, that's great, but advice is needed.

finat/aw.py
```diff
@@ -0,0 +1,158 @@
...
+            V[multiindex] = Literal(V[multiindex])
+
+        for i in range(0, 12, 2):
+            V[i, i] = Literal(1)
+
+        T = self.cell
+
+        # This bypasses the GEM wrapper.
+        that = numpy.array([T.compute_normalized_edge_tangent(i) for i in range(3)])
+        nhat = numpy.array([T.compute_normal(i) for i in range(3)])
```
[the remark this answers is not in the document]
rckirby: Could, but I will address this after more guidance on overall strategy (see above.)

finat/aw.py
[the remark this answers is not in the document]
rckirby: Again, glad to adopt these at the `gem` level, but I don't see `gem.Node` or its subclasses support `__setitem__`.  So I think I'm still stuck building up the `V` matrix in `numpy`?

finat/aw.py
```diff
@@ -0,0 +1,158 @@
...
+            Ghat_T = numpy.array([nhat[e, :], that[e, :]])
+
+            (alpha, beta) = Ghat_T @ JTJ @ that[e, :] / detJ
+
+            # Stuff into the right rows and columns.
+            (idx1, idx2) = (9 + 4*e + 1, 9 + 4*e + 3)
+            V[idx1, idx1-1] = Literal(-1) * alpha / beta
+            V[idx1, idx1] = Literal(1) / beta
+            V[idx2, idx2-1] = Literal(-1) * alpha / beta
+            V[idx2, idx2] = Literal(1) / beta
```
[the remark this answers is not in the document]
rckirby: I can do this after we figure out what that code should be with the `gem` <--> `numpy` business.

finat/fiat_elements.py
```diff
@@ -96,7 +96,9 @@ def basis_evaluation(self, order, ps, entity=None, coordinate_mapping=None):
                     assert np.allclose(table, 0.0)
                     exprs.append(gem.Zero(self.index_shape))
             if self.value_shape:
-                beta = self.get_indices()
+                # As above, this extent may be different from that advertised by the finat element.
+                assert len(self.get_indices()) == 1, "Was not expecting more than one index"
+                beta = (gem.Index(extent=self._element.space_dimension()), )
```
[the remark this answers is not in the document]
rckirby: I don't know the right answer here, but AW elements are intrinsically tensor-valued and this doesn't break.  Nevertheless, tests still pass with your change, which is simpler than what we had.

finat/mtw.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # This bypasses the GEM wrapper.
+        that = numpy.array([T.compute_normalized_edge_tangent(i) for i in range(3)])
+        nhat = numpy.array([T.compute_normal(i) for i in range(3)])
+
+        detJ = coordinate_mapping.detJ_at([1/3, 1/3])
+        J = coordinate_mapping.jacobian_at([1/3, 1/3])
+        J_np = numpy.array([[J[0, 0], J[0, 1]],
+                            [J[1, 0], J[1, 1]]])
+        JTJ = J_np.T @ J_np
```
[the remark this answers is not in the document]
rckirby: As above, we still have to stuff this into the final `V` matrix we return.  Coming up with a better way to do this could spawn a pervasive change (though not unwelcome) in all the zany elements.

finat/aw.py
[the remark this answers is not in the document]
miklos1: `ComponentTensor` basically represents a loop wrapped up as a tensor, all its entries must originate in the same rule of generation.  `ListTensor` is, on the other hand, a collection of scalar expressions that have (formally) nothing to do with each other.  Currently, there is very little in between: `Concatenate` nodes, which were introduced for a specific purpose and enjoy limited support.  Thus, it is not possible to get rid of the `ListTensor` at the end without substantial work in GEM and TSFC.

This does not mean, however, that one must use numpy up to the point of filling the `V` matrix.  Although sliced insertions won't work naïvely (e.g. `V[3:6, 3:6] = G` where `G` would be 3 x 3 GEM matrix), they are still doable, albeit a bit more complicated, for example:
```python
V[3:6, 3:6] = [[G[i, j]
                for j in range(3)]
               for i in range(3)]
```
or perhaps
```python
V[3:6, 3:6] = numpy.array([[G[i, j]
                            for j in range(3)]
                           for i in range(3)])
```
if necessary.

finat/aw.py
```diff
@@ -0,0 +1,158 @@
...
+        W[0, 2] = J[0, 1]*J[0, 1]
+        W[1, 0] = -1*J[1, 1]*J[1, 0]
+        W[1, 1] = J[1, 1]*J[0, 0] + J[0, 1]*J[1, 0]
+        W[1, 2] = -1*J[0, 1]*J[0, 0]
+        W[2, 0] = J[1, 0]*J[1, 0]
+        W[2, 1] = -2*J[1, 0]*J[0, 0]
+        W[2, 2] = J[0, 0]*J[0, 0]
+        W_check = W
+        # Put into the right rows and columns.
+        V[0:3, 0:3] = V[3:6, 3:6] = V[6:9, 6:9] = W_check
```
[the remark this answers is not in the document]
rckirby: We were re-scaling the degrees of freedom for conditioning but took that out because something was broken.  I guess we don't need the `W_check` for now.

finat/aw.py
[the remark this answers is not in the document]
rckirby: Since we're waiting to convert to newer gem features, I went ahead and did the refactoring.

finat/physically_mapped.py
rckirby: we can use it in a later PR once the tsfc changes land.  We need to make sure they are scaled/not scaled appropriately for this use (FIAT supports two kinds).

miklos1: Ok, so `reference_edge_tangets` is not used here in this PR, fine.

finat/aw.py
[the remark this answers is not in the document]
rckirby: Thanks, this is very helpful to my understanding.  Looking at the current code state, I'm still using numpy to do matrix manipulations (and we'll switch after the tsfc PR lands), and this will be more useful then once I've done more of the symbolic calculation at the gem rather than numpy level.

finat/aw.py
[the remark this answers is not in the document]
rckirby: We can do this as `gem.Literal` (or use further enhancements to physical geometry interface) once the tsfc PR has landed.

rckirby:
So @miklos1 and @wence- : are you happy with the current state until we get the tsfc/gem updates in place?

miklos1:
> So @miklos1 and @wence- : are you happy with the current state until we get the tsfc/gem updates in place?

Fine from me.

miklos1 (approved):

finat/physically_mapped.py
```diff
@@ -163,6 +201,14 @@ def jacobian_at(self, point):
...
         """
 
+    @abstractmethod
+    def detJ_at(self, point):
+        """The determinant of the jacobian of the physical coordinates at a point.
+
+        :arg point: The point in reference space to evaluate the Jacobian determinant.
+        :returns: A GEM expression for the Jacobian determinant.
+        """
+
```
miklos1, line 211: @rckirby: I made the same mistake earlier this week, but if you extends this interface, then you ought to merge the corresponding TSFC change (which provides the implementation) together with the FInAT change, otherwise Firedrake will break.

rckirby:
OK, I think we just did that since both have landed now.

PR #64, opened 2020-08-19
Fix age-old typo

miklos1:
This typo makes me wonder if these `#egg=` specifications are required or useful at all.


wence-:
Per https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support, it seems like they are possibly needed, but not necessary if not installing in editable mode directly from a repo (in firedrake-install we install in editable mode, but clone first).

wence- (approved):

PR #66, opened 2020-08-21
Interpolation onto a Vertex Only Mesh (Old)

ReubenHill:
Draft for changes needed to interpolate onto a Vertex Only Mesh from a suitable parent mesh.

Based on top of branch `dual_base_fiat` - see draft PR #57 . Base should change to master when that PR is complete (or if this diverges and stops being dependent on #57).

Linked PRs: https://github.com/firedrakeproject/firedrake/pull/1815 & https://github.com/firedrakeproject/tsfc/pull/226

### DIFF
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -53,13 +53,13 @@ class PointSingleton(AbstractPointSet):
         assert len(point.shape) == 1
         self.point = point
 
-    @property
+    @cached_property
     def points(self):
         # Make sure we conform to the expected (# of points, point dimension)
         # shape
         return self.point.reshape(1, -1)
 
-    @property
+    @cached_property
     def indices(self):
         return ()
 
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -69,12 +69,28 @@ class PointSingleton(AbstractPointSet):
 
 
 class UnknownPointSingleton(AbstractPointSet):
-    """A single unknown point."""
+    """A point set representing a single point with unknown location but known
+    `gem.Variable` expression.
+
+    Stored points are `numpy.nan`s to indicate that they are not known.
+
+    In the future this can be easily replaced with an equivalent for a whole
+    point set representing multiple points though storing the full array of
+    `numpy.nans` may become undesirable and require interface changes."""
 
     def __init__(self, point_expr):
+        """Build a PointSingleton from a gem expression for a single point.
+
+        :arg point_expr: A `gem.Variable` expression representing a single
+            point of shape (D,) where D is the dimension of the point.
+            Since this represents a single point and is a `gem.Variable` it
+            should have no free indices."""
         assert isinstance(point_expr, gem.Variable)
+        # Check we only have 1 point
+        assert point_expr.free_indices == ()
+        assert len(point_expr.shape) == 1
         self.expression = point_expr
-        self.points = numpy.full((1, 1), numpy.nan)
+        self.points = numpy.full(point_expr.shape, numpy.nan).reshape(1, -1)
 
     @cached_property
     def points(self):
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -68,6 +68,27 @@ class PointSingleton(AbstractPointSet):
         return gem.Literal(self.point)
 
 
+class UnknownPointSingleton(AbstractPointSet):
+    """A single unknown point."""
+
+    def __init__(self, point_expr):
+        assert isinstance(point_expr, gem.Variable)
+        self.expression = point_expr
+        self.points = numpy.full((1, 1), numpy.nan)
+
+    @cached_property
+    def points(self):
+        pass  # set at initialisation
+
+    @cached_property
+    def indices(self):
+        return self.expression.free_indices
+
+    @cached_property
+    def expression(self):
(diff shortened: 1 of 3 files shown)


finat/point_set.py
```diff
@@ -69,12 +69,28 @@ def expression(self):
...
+            point of shape (D,) where D is the dimension of the point.
+            Since this represents a single point and is a `gem.Variable` it
+            should have no free indices."""
         assert isinstance(point_expr, gem.Variable)
+        # Check we only have 1 point
+        assert point_expr.free_indices == ()
+        assert len(point_expr.shape) == 1
         self.expression = point_expr
-        self.points = numpy.full((1, 1), numpy.nan)
+        self.points = numpy.full(point_expr.shape, numpy.nan).reshape(1, -1)
```
ReubenHill, line 93: If anyone can think of a nicer way of doing this without having to actually allocate memory I'm all ears. As I say in the comment, extending this to a full `PointSet` of unknown points should be pretty easy but it would be daft to be carrying around an array of `NaN`s

ReubenHill: Perhaps the `AbstractPointSet` needs to be adjusted to not require a cached `numpy.array` as the `points` property

finat/point_set.py
```diff
@@ -68,6 +68,43 @@ def expression(self):
         return gem.Literal(self.point)
 
 
+class UnknownPointSingleton(AbstractPointSet):
+    """A point set representing a single point with unknown location but known
+    `gem.Variable` expression.
```
wence-, line 73: This is not the right way to link things, if you want fixed width, you need two ` at each end.

finat/point_set.py
```diff
@@ -68,6 +68,43 @@ def expression(self):
...
+
+    Stored points are `numpy.nan`s to indicate that they are not known.
+
+    In the future this can be easily replaced with an equivalent for a whole
+    point set representing multiple points though storing the full array of
+    `numpy.nans` may become undesirable and require interface changes."""
```
wence-, line 79: This is a comment, not a docstring.

finat/point_set.py
wence-: When do you want to be able to realise that the `points` property is invalid? I would `raise ValueError("Unknown points do not have a point array")` in the property.

ReubenHill: Yes I tried that, the problem is that `points` is used all over the codebase to check for things like length matching.

finat/quadrature_element.py
```diff
@@ -99,18 +104,24 @@ def point_evaluation(self, order, refcoords, entity=None):
     def dual_basis(self):
         # Point evaluations at each quadrature point with no derivatives
         duals = []
-        for pt in self._rule.point_set.points:
-            point_set = PointSingleton(pt)
+        if isinstance(self._rule.point_set, UnknownPointSingleton):
+            point_set = self._rule.point_set
```
wence-, line 108: I think this code is hamstrung by the lack of actual data structure for the duals.

I think you can avoid this special casing if you change the iteration over the points to be something like:


```
points = gem.ComponentTensor(self._rule.point_set.expression, self._rule.indices)
for idx in numpy.ndindex(points.shape):
    pt = gem.partial_indexed(points, idx)
    duals.append(pt ...)
```
Or does the duals "data structure" need something different there?

ReubenHill:
@wence- before you go review-crazy, note that I'm currently working on hacking interpolation onto a Vertex Only Mesh onto `master` using the existing dual evaluation methods. Once merged, I merge any relevant commits from here into #57 and then deal with all the dual evaluation work in one go.

finat/point_set.py
wence-: If you need something that advertises the "meta-data" part of the numpy array just make a mock object.

wence-:
> @wence- before you go review-crazy, note that I'm currently working on hacking interpolation onto a Vertex Only Mesh onto `master` using the existing dual evaluation methods. Once merged, I merge any relevant commits from here into #57 and then deal with all the dual evaluation work in one go.

OK, thanks. We should do some pen-and-paper data structure design for that...

ReubenHill:
> > @wence- before you go review-crazy, note that I'm currently working on hacking interpolation onto a Vertex Only Mesh onto `master` using the existing dual evaluation methods. Once merged, I merge any relevant commits from here into #57 and then deal with all the dual evaluation work in one go.
> 
> OK, thanks. We should do some pen-and-paper data structure design for that...

I agree. It's a bit of a mess at present. Whenever we do do that, I believe @pbrubeck is up for helping.

finat/point_set.py
ReubenHill: good idea

miklos1:
> I don't have time for a detailed explanation at the moment, but I feel like `UnknownPointSet` tries to accomplish via `tsfc.fem.PointSetContext` merely what has already been done through `tsfc.fem.GemPointContext`.

In fact, I have said [this before](https://github.com/firedrakeproject/tsfc/pull/226#issuecomment-679866781).

ReubenHill:
Replaced by #81

ReubenHill:
Note all relevant comments have been moved to #81

PR #68, opened 2020-08-24
Use install_requires instead of requirements-ext.txt

miklos1:
This ought to fix [this failure](https://travis-ci.org/github/firedrakeproject/tsfc/builds/720705935) on Travis CI.  That TSFC build is failing, because FInAT dependencies from `requirements-git.txt` are not getting installed in the TSFC test environment.  (`firedrake-install` manually takes care of this.)

Listing dependencies instead in `install_requires` make sure that `pip` always installs them when these packages are needed.

### DIFF
--- a/requirements-ext.txt
+++ b/requirements-ext.txt
@@ -1,3 +1,0 @@
-numpy>=1.16
-symengine
-sympy
--- a/requirements-ext.txt
+++ b/requirements-ext.txt
@@ -1,0 +1,3 @@
+numpy>=1.16
+symengine
+sympy
--- a/requirements-git.txt
+++ b/requirements-git.txt
@@ -1,0 +1,2 @@
+git+https://github.com/firedrakeproject/fiat.git#egg=fiat
+git+https://github.com/firedrakeproject/tsfc.git#egg=tsfc
--- a/requirements-git.txt
+++ b/requirements-git.txt
@@ -1,2 +1,0 @@
-git+https://github.com/firedrakeproject/fiat.git#egg=fiat
-git+https://github.com/firedrakeproject/tsfc.git#egg=tsfc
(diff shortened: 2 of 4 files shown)


rckirby (dismissed):

miklos1:
This breaks `firedrake-install` at this point.  Looking into it...

miklos1:
Testing this now...

I expect this to work, but with this we now have a double accounting of direct dependencies.  Not ideal...

miklos1:
Maybe it's enough to empty out `requirements-ext.txt`...

miklos1:
> Maybe it's enough to empty out `requirements-ext.txt`...

Nope, we apparently use `--no-deps` in `firedrake-install`... this is horrible.

dham:
> > Maybe it's enough to empty out `requirements-ext.txt`...
> 
> Nope, we apparently use `--no-deps` in `firedrake-install`... this is horrible.

We don't use `--no-deps` for all packages so maybe we can stop using it for FInAT

rckirby:
I'm glad to do support whatever makes sense.  It seems like turning off `no-deps` for FInAT in firedrake install makes sense, but I don't know enough about the build system to understand possible implications of this.

miklos1:
> We don't use `--no-deps` for all packages so maybe we can stop using it for FInAT

We should do a more thorough fix, I will write an issue about it.

But now I need to get this merged to unbreak things.  Can someone please tick approve?

rckirby:
Done

miklos1:
> Done

Thanks.  This is a temporary fix, but this shall unbreak CI for TSFC.

miklos1:
> > We don't use `--no-deps` for all packages so maybe we can stop using it for FInAT
> 
> We should do a more thorough fix, I will write an issue about it.

See firedrakeproject/firedrake#1821.

PR #69, opened 2020-08-28
Fix tabulation for zany elements

wence-:
We do need to obey index_shape, but we need to generate indices whose
extents match index_shape.


wence-:
Someone didn't actually check that the changes they merged were right...

pefarrell (approved):
This fixes the MTW convergence tests (that we have not yet made a PR to contribute to firedrake).

PR #73, opened 2021-02-04
Support useful testing element

ReubenHill:
Adds support a new FIAT element with integral moment nodes and point evaluation nodes on an interval cell which is very useful for sanity checking. Name is very much up for negotiation!

Linked PRs:

https://github.com/FEniCS/fiat/pull/64
https://github.com/FEniCS/ufl/pull/47
https://github.com/firedrakeproject/tsfc/pull/239


ReubenHill:
I've decided to move this into the test itself and not create a public element (see https://github.com/firedrakeproject/firedrake/pull/1971) so will close this.

PR #76, opened 2021-03-10
Make symengine an optional dependency

wence-:


wence-:
ARM wheels are not available yet.

miklos1 (approved):

PR #79, opened 2021-04-28
Fix PointSingleton and make PointSet work for single points

### LINKED ISSUE #78: Can't create PointSet.expression for when have empty points list
Running this code results in the following error
```python
from finat.point_set import PointSet
import numpy as np

# This is a valid point set at the moment - e.g. for P0DG on a vertex
# which, in FIAT, defines the reference cell as being at an empty tuple
# point and so gives a dual.pt_dict of {(): [(1.0, ())]}.
# This can be seen happening in tsfc/driver.py lines 362-369.
points = [()]
# Can create a PointSet
ps = PointSet(points)
# Have 0 points
assert ps.points.shape == np.array([[]]).shape
# This errors when trying to make expression using
# gem.partial_indexed(gem.Literal(self.points), self.indices)
ps.expression
```

```
  File "FInAT_issue.py", line 15, in <module>
    ps.expression
  File "/Users/rwh10/firedrake/src/tsfc/gem/utils.py", line 19, in __get__
    obj.__dict__[self.__name__] = result = self.fget(obj)
  File "/Users/rwh10/firedrake/src/FInAT/finat/point_set.py", line 75, in expression
    return gem.partial_indexed(gem.Literal(self.points), self.indices)
  File "/Users/rwh10/firedrake/src/tsfc/gem/gem.py", line 882, in partial_indexed
    return ComponentTensor(
  File "/Users/rwh10/firedrake/src/tsfc/gem/gem.py", line 48, in __call__
    obj = super(NodeMeta, self).__call__(*args, **kwargs)
  File "/Users/rwh10/firedrake/src/tsfc/gem/gem.py", line 654, in __new__
    assert all(shape)
```

This clearly raises some questions about what the `PointSet.expression` should be, nevertheless the fact that you can make such a `PointSet` suggests the `.expression` should be accessible.

ReubenHill:
Fixes the `PointSingleton` interface not matching the advertised `AbstractPointSet` interface. Also attempts to fix #78

### DIFF
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -72,6 +72,9 @@ class PointSet(AbstractPointSet):
 
     @cached_property
     def expression(self):
+        npoints, _ = self.points.shape
+        if npoints <= 1:
+            return gem.Literal(self.points[0])
         return gem.partial_indexed(gem.Literal(self.points), self.indices)
 
     def almost_equal(self, other, tolerance=1e-12):
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -37,21 +37,21 @@ class PointSingleton(AbstractPointSet):
     """Just a single point."""
 
     def __init__(self, point):
-        point = numpy.asarray(point)
-        assert len(point.shape) == 1
+        point = numpy.asarray(point).reshape(1, -1)
+        assert point.shape[0] == 1
         self.point = point
 
     @property
     def points(self):
-        return self.point.reshape(1, -1)
+        return self.point
 
     @property
     def indices(self):
-        return ()
+        return (gem.Index(extent=len(self.points)),)
 
     @cached_property
     def expression(self):
-        return gem.Literal(self.point)
+        return gem.Literal(self.point[0])
 
 
 class PointSet(AbstractPointSet):


finat/point_set.py
```diff
@@ -37,21 +37,21 @@ class PointSingleton(AbstractPointSet):
...
 
     @property
     def points(self):
-        return self.point.reshape(1, -1)
+        return self.point
 
     @property
     def indices(self):
-        return ()
+        return (gem.Index(extent=len(self.points)),)
```
miklos1, line 50: I don't see the point of these changes to `PointSingleton`.  What purpose does this class still have after these changes over just using `PointSet` with a single point?

finat/point_set.py
```diff
@@ -72,6 +72,9 @@ def indices(self):
 
     @cached_property
     def expression(self):
+        npoints, _ = self.points.shape
+        if npoints <= 1:
+            return gem.Literal(self.points[0])
```
miklos1, line 77: Now this one special cases `PointSet` to behave like `PointSingleton` in case of the single point.  I kind of suspect this is going to cause trouble because the index that is there in the general case, will now be missing in a particular case...

miklos1: > Now this one special cases `PointSet` to behave like `PointSingleton`

like the original `PointSingleton`, before the changes that are being proposed here.

miklos1:
I don't like this, and if #78 truly needs fixing, perhaps try relaxing the assertion in the `ComponentTensor` constructor, and see if that helps.

finat/point_set.py
ReubenHill: `PointSingleton` is used throughout the code base and often dispatched on or type checked for. I suppose I could go through the code base and see if I can replace it everywhere? Or add a deprecation warning to its `__init__`?

ReubenHill:
> I don't like this, and if #78 truly needs fixing, perhaps try relaxing the assertion in the `ComponentTensor` constructor, and see if that helps.

What exactly don't you like? The changes to fix `PointSingleton`'s behaviour to match the advertised API or the changes made to fix #78 ?

finat/point_set.py
ReubenHill: Good point

finat/point_set.py
ReubenHill: Perhaps this ought to return `((),)`?

ReubenHill: Ah actually that only makes sense if `()` is a valid `gem.Index`. I may have misunderstood what counts as a valid index when I made these changes

miklos1:
> What exactly don't you like? The changes to fix `PointSingleton`'s behaviour to match the advertised API or the changes made to fix #78 ?

I just had a look at the claim that `PointSingleton` not match the interface advertised by `AbstractPointSet`, and I don't find this to be true.  What promise do you think is broken?

I dislike both the changes to `PointSingleton` and to `PointSet`.  I might be convinced otherwise, or just be wrong, but so far I suspect that fixing #78 is probably best done by relaxing the `ComponentTensor` constructor assertion.

finat/point_set.py
miklos1: > `PointSingleton` is used throughout the code base and often dispatched on or type checked for. I suppose I could go through the code base and see if I can replace it everywhere? Or add a deprecation warning to its `__init__`?

This would be a logical next clean-up step, if the changes suggested here to `PointSingleton` were reasonable, which I strongly doubt.

ReubenHill:
I think you are correct and I had misunderstood the API of PointSingleton (in particular the intended link between `point` and `points`. I've made a new PR (#80) where I try to make this clear in case anyone else gets confused and tries to fix what isn't broken.

PR #80, opened 2021-04-29
Add docstrings to PointSingleton and PointSet

ReubenHill:
Aims to avoid any ambiguity or confusion about the expected shape of the `__init__` arguments and generally avoid the silly confusion that led to me making 3b19867bb15cdcac1b7accea73c38108f0fe46b0

### DIFF
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -9,7 +9,12 @@ from gem.utils import cached_property
 
 class AbstractPointSet(metaclass=ABCMeta):
     """A way of specifying a known set of points, perhaps with some
-    (tensor) structure."""
+    (tensor) structure.
+
+    Points, when stored, have shape point_set_shape + (point_dimension,)
+    where point_set_shape is () for scalar, (N,) for N element vector,
+    (N, M) for N x M matrix etc.
+    """
 
     @abstractproperty
     def points(self):
@@ -34,15 +39,24 @@ class AbstractPointSet(metaclass=ABCMeta):
 
 
 class PointSingleton(AbstractPointSet):
-    """Just a single point."""
+    """A point set representing a single point.
+
+    These have a `gem.Literal` expression and no indices."""
 
     def __init__(self, point):
+        """Build a PointSingleton from a single point.
+
+        :arg point: A single point of shape (D,) where D is the dimension of
+            the point."""
         point = numpy.asarray(point)
+        # 1 point ought to be a 1D array - see docstring above and points method
         assert len(point.shape) == 1
         self.point = point
 
     @property
     def points(self):
+        # Make sure we conform to the expected (# of points, point dimension)
+        # shape
         return self.point.reshape(1, -1)
 
     @property
@@ -55,9 +69,14 @@ class PointSingleton(AbstractPointSet):
 
 
 class PointSet(AbstractPointSet):
-    """A basic point set with no internal structure."""
+    """A basic point set with no internal structure representing a vector of
+    points."""
 
     def __init__(self, points):
+        """Build a PointSet from a vector of points
+
+        :arg points: A vector of N points of shape (N, D) where D is the
+            dimension of each point."""
         points = numpy.asarray(points)
         assert len(points.shape) == 2
         self.points = points


miklos1:
The idea is that the set of points could in principled be arranged in a tensor shape.  You may have 4 x 4 x 4 points on a 3D tensor product cell, which is distinguished from having 64 flat points, preserving the structure that is need for sum factorisation etc.

So there is a shape corresponding to the shape of the point set, and then there is the shape of the points, the latter is always rank one, as a point is always a vector.

For `PointSingleton` the point set shape is always `()` (i.e. scalar). `PointSet` represents a flat array of points, and other classes do more fancy stuff such as tensor product point set.

The `.points` property is used with things like FIAT that always expect a flat array of points, so that property always flattens dimensions to a vector of points.

finat/point_set.py
```diff
@@ -34,15 +34,23 @@ def expression(self):
 
 
 class PointSingleton(AbstractPointSet):
-    """Just a single point."""
+    """A point set representing a single point. It therefore has no indices an
+    a `gem.Literal` expression."""
```
miklos1, line 38: Second sentence not quite clear, "an a" certainly wrong.

finat/point_set.py
```diff
@@ -34,15 +34,23 @@ def expression(self):
...
 
     def __init__(self, point):
+        """Build a PointSingleton from a single point.
+
+        :arg point: A single point which, once turned into a numpy array, has
+            shape (point dimension,)."""
```
miklos1, line 44: OK

finat/point_set.py
```diff
@@ -34,15 +34,23 @@ def expression(self):
...
         point = numpy.asarray(point)
+        # Ensure we only have 1 point
         assert len(point.shape) == 1
```
miklos1, line 47: The assertion is a bit stronger than the comment: the point set shape is basically scalar `()`, rather than a single- element vector `(1,)` or 1 x 1 matrix `(1, 1)`.  So `point.shape` is merely the shape of the point, i.e. a vector.

finat/point_set.py
```diff
@@ -34,15 +34,23 @@ def expression(self):
...
         self.point = point
 
     @property
     def points(self):
+        # Make sure we conform to the expected (# of points, point dimension)
+        # shape
```
miklos1, line 53: OK

finat/point_set.py
```diff
@@ -58,6 +66,10 @@ class PointSet(AbstractPointSet):
     """A basic point set with no internal structure."""
 
     def __init__(self, points):
+        """Build a PointSet from a set of points
+
+        :arg points: A set of points which, once turned into a numpy array,
+            have shape (# of points, point dimension)."""
```
miklos1, line 72: okay

finat/point_set.py
ReubenHill: I'm not sure I follow - the `point` has shape `(point dimension,)` not `()` here and the assertion checks that no?

finat/point_set.py
ReubenHill: Updated

finat/point_set.py
ReubenHill: Updated comment

ReubenHill: (The thing that I found contentious was that I couldn't initialise a PointSingleton with something of shape `(point dimension, 1)` and have it just work - you have to squeeze down the number of dimensions such that you have something of shape `(point dimension, )`)

miklos1: The shape of the internal array is always `point_set_shape + (point_dimension,)` in the general case, and in case of `PointSingleton` `point_set_shape == ()`, while in case of `PointSet` `point_set_shape = (npoints,)`.

ReubenHill: Okay I've tried to make clarify this in 2066f381b157db00075124e58626c83c00fed7e1 . A `PointSet` deals with a vector of points specifically whilst a `PointSingleton` is a single scalar point.

I suppose it might be nice one day to be able to make an `NxMxL` `TensorPointSet` of points with dimension `D` directly from a numpy `ndarray` of shape `(N, M, L) + (D) = (N, M, L, D)`

wence-: Sure, although normally you're carrying around pointsets by this point, no?

miklos1: Yes, normally you wouldn't want to construct the outer product ndarray of a TensorPointSet, but rather just tabulate the factor elements of a tensor product element on the factor point sets of the tensor point set...

finat/point_set.py
```diff
@@ -34,15 +39,15 @@ def expression(self):
 
 
 class PointSingleton(AbstractPointSet):
-    """A point set representing a single point.
+    """A point set representing a single scalar point.
```
miklos1, line 42: I find the term "scalar point" confusing, that sounds like a 1-dimensional point or something.

finat/point_set.py
```diff
@@ -34,15 +39,15 @@ def expression(self):
...
 
     These have a `gem.Literal` expression and no indices."""
 
     def __init__(self, point):
         """Build a PointSingleton from a single point.
 
-        :arg point: A single point which, once turned into a numpy array, has
-            shape (point dimension,)."""
+        :arg point: A single scalar point of shape (D,) where D is the
+            dimension of the point."""
```
miklos1, line 50: Same here with "scalar point", I would just drop the word scalar.

miklos1 (commented):
Aside from "scalar point", I think this is fine.

finat/point_set.py
ReubenHill: Removed in 8333ae7bcca2fd963d3b952713d66be8b8026749

finat/point_set.py
ReubenHill: Removed in 8333ae7bcca2fd963d3b952713d66be8b8026749

ReubenHill:
@wence- I've squashed into one commit and passed the Miklos test

miklos1 (approved):
This is fine now.

ReubenHill:
Thanks @miklos1 !

PR #81, opened 2021-05-07
Runtime point set for use with QuadratureElement

ReubenHill:
Adds a new `UnknownPointSingleton` which represents a single point which is specified at run time. 

`UnknownPointSingleton`s are typically intended for use as the point set in the rule of a `QuadratureElement`: the `QuadratureElement` acts as a target element on the reference cell when performing dual basis evaluation on some other element for run time specified points. A typical use case is when performing dual evaluation across different elements on different domains. By only specifying an `UnknownPointSingleton` (rather than a full `UnknownPointSet`), only runtime evaluation at single points can be supported at present. 

The `QuadratureElement` API has been suitably adjusted and a compatibility function added.

These changes are needed to allow interpolation from an arbitrary function function space onto a P0DG function space on a Firedrake `VertexOnlyMesh`

This replaces #66 and sits on top of master, avoiding dependency on #57 (dual evaluation which needs lots of work).

Linked PRs
 - https://github.com/firedrakeproject/firedrake/pull/2059
 - https://github.com/firedrakeproject/tsfc/pull/247

Generated Issues
 - ~#83~
 - #84 

todo:
 - [x] ~Add comment about array of NaNs being a hack to support FIAT dual evaluation and that, when FInAT dual evaluation is implemented, we will raise an exception when asking for the `points` of an `UnknownPointSingleton` and will disallow creation of a `fiat_equivalent` for a quadrature element which uses an `UnknownPointSingleton`~ See dbc58367796bc6ec4605f65e475946f4a58b423a

### DIFF
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -18,8 +18,8 @@ class AbstractPointSet(metaclass=ABCMeta):
 
     @abstractproperty
     def points(self):
-        """A flattened numpy array of points with shape
-        (# of points, point dimension)."""
+        """A flattened numpy array of points or ``UnknownPointsArray``
+        object with shape (# of points, point dimension)."""
 
     @property
     def dimension(self):
@@ -41,7 +41,7 @@ class AbstractPointSet(metaclass=ABCMeta):
 class PointSingleton(AbstractPointSet):
     """A point set representing a single point.
 
-    These have a `gem.Literal` expression and no indices."""
+    These have a ``gem.Literal`` expression and no indices."""
 
     def __init__(self, point):
         """Build a PointSingleton from a single point.
@@ -64,6 +64,60 @@ class PointSingleton(AbstractPointSet):
     @cached_property
     def expression(self):
         return gem.Literal(self.point)
+
+
+class UnknownPointsArray():
+    """A placeholder for a set of unknown points with appropriate length
+    and size but without indexable values. For use with
+    :class:`AbstractPointSet`s whose points are not known at compile
+    time."""
+    def __init__(self, shape):
+        """
+        :arg shape: The shape of the unknown set of N points of shape
+            (N, D) where D is the dimension of each point.
+        """
+        assert len(shape) == 2
+        self.shape = shape
+
+    def __len__(self):
+        return self.shape[0]
+
+
+class UnknownPointSet(AbstractPointSet):
+    """A point set representing a vector of points with unknown
+    locations but known ``gem.Variable`` expression.
+
+    The ``.points`` property is an `UnknownPointsArray` object with
+    shape (N, D) where N is the number of points and D is their
+    dimension.
+
+    The ``.expression`` property is a derived `gem.partial_indexed` with
+    shape (D,) and free indices for the points N."""
+
+    def __init__(self, points_expr):
+        """Build a PointSingleton from a gem expression for a single point.
+
+        :arg points_expr: A ``gem.Variable`` expression representing a
+            vector of N points in D dimensions. Should have shape (N, D)
+            and no free indices. For runtime tabulation the variable
+            name should begin with \'rt_:\'."""
+        assert isinstance(points_expr, gem.Variable)
+        assert points_expr.free_indices == ()
+        assert len(points_expr.shape) == 2
+        self._points_expr = points_expr
+
+    @cached_property
+    def points(self):
+        return UnknownPointsArray(self._points_expr.shape)
+
+    @cached_property
+    def indices(self):
+        N, _ = self._points_expr.shape
+        return (gem.Index(extent=N),)
+
+    @cached_property
+    def expression(self):
+        return gem.partial_indexed(self._points_expr, self.indices)
 
 
 class PointSet(AbstractPointSet):
--- a/finat/point_set.py
+++ b/finat/point_set.py
@@ -53,15 +53,13 @@ class PointSingleton(AbstractPointSet):
         assert len(point.shape) == 1
         self.point = point
 
-    @property
+    @cached_property
     def points(self):
         # Make sure we conform to the expected (# of points, point dimension)
         # shape
         return self.point.reshape(1, -1)
 
-    @property
-    def indices(self):
-        return ()
+    indices = ()
 
     @cached_property
     def expression(self):
--- a/finat/quadrature_element.py
+++ b/finat/quadrature_element.py
@@ -12,17 +12,38 @@ from finat.finiteelementbase import FiniteElementBase
 from finat.quadrature import make_quadrature, AbstractQuadratureRule
 
 
+def make_quadrature_element(fiat_ref_cell, degree, scheme="default"):
+    """Construct a :class:`QuadratureElement` from a given a reference
+    element, degree and scheme.
+
+    :param fiat_ref_cell: The FIAT reference cell to build the
+        :class:`QuadratureElement` on.
+    :param degree: The degree of polynomial that the rule should
+        integrate exactly.
+    :param scheme: The quadrature scheme to use - e.g. "default",
+        "canonical" or "KMV".
+    :returns: The appropriate :class:`QuadratureElement`
+    """
+    rule = make_quadrature(fiat_ref_cell, degree, scheme)
+    return QuadratureElement(fiat_ref_cell, rule)
+
+
 class QuadratureElement(FiniteElementBase):
     """A set of quadrature points pretending to be a finite element."""
 
-    def __init__(self, cell, degree, scheme="default", rule=None):
-        self.cell = cell
-        if rule is not None:
-            if not isinstance(rule, AbstractQuadratureRule):
-                raise TypeError("rule is not an AbstractQuadratureRule")
-            self._rule = rule
-        else:
-            self._rule = make_quadrature(cell, degree, scheme)
+    def __init__(self, fiat_ref_cell, rule):
+        """Construct a :class:`QuadratureElement`.
+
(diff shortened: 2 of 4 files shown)


finat/point_set.py
```diff
@@ -53,13 +53,13 @@ def __init__(self, point):
...
+    @cached_property
     def points(self):
         # Make sure we conform to the expected (# of points, point dimension)
         # shape
         return self.point.reshape(1, -1)
 
-    @property
+    @cached_property
     def indices(self):
         return ()
```
wence-, line 64: Any particular reason for these changes?

wence-: I guess no reason not to, although note that you can replace this complicated indices thing by just `indices = ()`.

finat/point_set.py
```diff
@@ -68,6 +68,43 @@ def expression(self):
...
 
+class UnknownPointSingleton(AbstractPointSet):
+    """A point set representing a single point with unknown location but known
+    `gem.Variable` expression.
+
+    Stored points are `numpy.nan`s to indicate that they are not known.
+
+    In the future this can be easily replaced with an equivalent for a whole
+    point set representing multiple points though storing the full array of
+    `numpy.nans` may become undesirable and require interface changes."""
```
wence-, line 79: The comments I made on the old PR still apply here.

ReubenHill:
@miklos1 to carry on the conversation from the old PR

> > I don't have time for a detailed explanation at the moment, but I feel like `UnknownPointSet` tries to accomplish via `tsfc.fem.PointSetContext` merely what has already been done through `tsfc.fem.GemPointContext`.
> 
> In fact, I have said [this before](https://github.com/firedrakeproject/tsfc/pull/226#issuecomment-679866781).

The `UnknownPointSingleton` gets translated into a `tsfc.fem.GemPointContext` in TSFC. Within FInAT, however, it still seems to make sense that the set of points that are represented be held in an `AbstractPointSet` of some kind (it's a set of points after all).

finat/point_set.py
```diff
@@ -68,6 +68,43 @@ def expression(self):
...
+        :arg point_expr: A `gem.Variable` expression representing a single
+            point of shape (D,) where D is the dimension of the point.
+            Since this represents a single point and is a `gem.Variable` it
+            should have no free indices."""
+        assert isinstance(point_expr, gem.Variable)
+        # Check we only have 1 point
+        assert point_expr.free_indices == ()
+        assert len(point_expr.shape) == 1
+        self.expression = point_expr
+        self.points = numpy.full(point_expr.shape, numpy.nan).reshape(1, -1)
```
ReubenHill, line 93: Idea from @wence- 
> If you need something that advertises the "meta-data" part of the numpy array just make a mock object.

finat/point_set.py
ReubenHill: Note old comment
> This is a comment, not a docstring.

finat/quadrature_element.py
```diff
@@ -9,15 +9,20 @@
...
-    def __init__(self, cell, degree, scheme="default"):
+    def __init__(self, cell, degree, scheme="default", rule=None):
         self.cell = cell
-        self._rule = make_quadrature(cell, degree, scheme)
+        if rule is not None:
+            if not isinstance(rule, AbstractQuadratureRule):
+                raise TypeError("rule is not an AbstractQuadratureRule")
+            self._rule = rule
+        else:
+            self._rule = make_quadrature(cell, degree, scheme)
```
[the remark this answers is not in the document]
miklos1: Indeed, this is a bad API.  I agree with @wence-'s suggestion on how to go forward if this needs to be done.

miklos1:
> > > I don't have time for a detailed explanation at the moment, but I feel like `UnknownPointSet` tries to accomplish via `tsfc.fem.PointSetContext` merely what has already been done through `tsfc.fem.GemPointContext`.
>
> The `UnknownPointSingleton` gets translated into a `tsfc.fem.GemPointContext` in TSFC. Within FInAT, however, it still seems to make sense that the set of points that are represented be held in an `AbstractPointSet` of some kind (it's a set of points after all).

It seems to me that the immediate reason you want to do this, is that you want to use a `QuadratureElement` for run-time point evaluations.  But why do you want to do that?

finat/point_set.py
ReubenHill: These are `@abstractproperty` in `class AbstractPointSet(metaclass=ABCMeta)` so you have to write
```python
    @cached_property
    def indices(self):
        pass
```
if you set it at initialisation so it all seems much-of-a-muchness

miklos1:
Perhaps there are good reasons for doing things this way, but so far I have not heard those good reasons.

> Because the dual basis to a quadrature element is point evaluation so it makes sense to use it.

This is not unique to quadrature elements; Langrange, discontinuous Lagrange, Serendipity, dPC, etc. also have only point evaluation nodes, which means their dual bases are just point evaluations.  But the key difference here, is that thus far all these elements have nodes whose reference coordinates are compile-time known, and now you are introducing quadrature elements with quadrature points that are only known at run time.

> When we scale up to higher dimensions (e.g. runtime quadrature of a line or the intersection of two elements) then a runtime `QuadratureElement` gets us what we want.

Can you elaborate this?

> Why bother having two interfaces rather than one?
> Do you propose an alternative?

So when I was active, there were three ways to compile UFL:
1. Form compilation with `tsfc.driver.compile_form`
2. Interpolation with `tsfc.compile_expression_at_points`, which was since extended to support a wider range of element types to interpolate into, and is now called `tsfc.compile_expression_dual_evaluation`
3. Point evaluation with `firedrake.pointeval_utils.compile_element` which generates code for point evaluations at points only known at run time.

Now, it seems that another way of doing point evaluation is being squeezed into point 2, by making `PointSet` no longer a set of compile-time known points, but potentially a placeholder for points not known at compile-time, and then scattered around the interpolation pipeline we branch based on whether 2 or 3 is actually being done.  How is this design not more convoluted than it was before?

dham:
> So when I was active, there were three ways to compile UFL:
> 
> 1. Form compilation with `tsfc.driver.compile_form`
> 2. Interpolation with `tsfc.compile_expression_at_points`, which was since extended to support a wider range of element types to interpolate into, and is now called `tsfc.compile_expression_dual_evaluation`
> 3. Point evaluation with `firedrake.pointeval_utils.compile_element` which generates code for point evaluations at points only known at run time.
> 
> Now, it seems that another way of doing point evaluation is being squeezed into point 2, by making `PointSet` no longer a set of compile-time known points, but potentially a placeholder for points not known at compile-time, and then scattered around the interpolation pipeline we branch based on whether 2 or 3 is actually being done. How is this design not more convoluted than it was before?

Using an element with a runtime dual space you can see 3 as a special case of 2. So this is moving towards just having two pathways, compile form or compile expression. This isn't all the way there yet which is why 3 is still there, but it will eventually go.

finat/quadrature_element.py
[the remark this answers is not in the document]
ReubenHill: Changed in da868b83c18049a7cbbfc7e4547123afa60c567b

finat/point_set.py
ReubenHill: So I've tried adding this
```python
class UnknownPoints():
    """An object that behaves like a set of known points (with
    appropriate length and size) but which is unknown. For use with
    :class:`AbstractPointSet`s whos points are not known at compile
    time. If indexed, return numpy.nans."""
    def __init__(self, shape):
        """
        :arg shape: The shape of the unknown set of N points of shape
            (N, D) where D is the dimension of each point.
        """
        assert len(shape) == 2
        self.shape = shape

    def __len__(self):
        return self.shape[0]
```
the trouble is that `ps.points` needs to be iterable and subscriptable when passed to, for example, FIAT on [this line](https://github.com/FInAT/FInAT/blob/1e0e97ddce99cd3ec8f09169eea8f35980749b42/finat/quadrature_element.py#L93).
so I either need to add special cases (yuck) or do something like the below (but with correct indexing) 
```python
    def __getitem__(self, idx):
        return numpy.full(self.shape[1], numpy.nan)
```

wence-: I guess this pretty much suggests that something is not right in the design. I think that one can't plausibly construct a fiat equivalent to these things. 

But just slap an array of nans in there for now.

ReubenHill: Okay I will leave as is for now. Have added a todo comment and will make an issue

ReubenHill: Created issue #83 .

finat/point_set.py
```diff
@@ -53,21 +53,74 @@ def __init__(self, point):
...
     @cached_property
     def expression(self):
         return gem.Literal(self.point)
 
 
+class UnknownPointsArray():
+    """A mock object that behaves like a set of known points (with
+    appropriate length and size) but without indexable values. For use
+    with :class:`AbstractPointSet`s who's points are not known at
+    compile time."""
```
ReubenHill, line 73: Rephrase: A placeholder for an array of unknown points

finat/point_set.py
```diff
@@ -53,21 +53,74 @@ def __init__(self, point):
...
+            (N, D) where D is the dimension of each point.
+        """
+        assert len(shape) == 2
+        self.shape = shape
+
+    def __len__(self):
+        return self.shape[0]
+
+
+class UnknownPointSingleton(AbstractPointSet):
```
ReubenHill, line 86: Change to `UnknownPointSet`

ReubenHill:
All comments addressed and all tests passing - ready for re review.

dham (approved):

PR #87, opened 2021-06-07
orientations: define entity_permutations() method for basic finite elements

ksagiyam:
This PR is to add entity_permutations `method` to basic finite FInAT elements (such as `TensorProductElement`) that defines a DoF permutation (to be used in global vector assembly) for each possible entity orientation for each entity for each dimension.
`entity_permutations` returns `None` if it is not yet implemented for that element, so the problem solving environment must handle this correctly if it ever tries to use this method.

Requires:
https://github.com/firedrakeproject/fiat/pull/25

### DIFF
--- a/finat/cube.py
+++ b/finat/cube.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, print_function, division
 
 from FIAT.reference_element import UFCHexahedron, UFCQuadrilateral
-from FIAT.reference_element import compute_unflattening_map, flatten_entities
+from FIAT.reference_element import compute_unflattening_map, flatten_entities, flatten_permutations
 from FIAT.tensor_product import FlattenedDimensions as FIAT_FlattenedDimensions
 
 from gem.utils import cached_property
@@ -48,6 +48,10 @@ class FlattenedDimensions(FiniteElementBase):
 
     def entity_dofs(self):
         return self._entity_dofs
+
+    @cached_property
+    def entity_permutations(self):
+        return flatten_permutations(self.product.entity_permutations)
 
     def space_dimension(self):
         return self.product.space_dimension()
--- a/finat/discontinuous.py
+++ b/finat/discontinuous.py
@@ -36,6 +36,16 @@ class DiscontinuousElement(FiniteElementBase):
     def entity_dofs(self):
         return self._entity_dofs
 
+    @cached_property
+    def entity_permutations(self):
+        # Return entity_permutations of the base finite element if it only
+        # has cell degrees of freedom; otherwise entity_permutations is not
+        # yet implemented for DiscontinuousElement.
+        if self.element.entity_dofs() == self.element.entity_closure_dofs():
+            return self.element.entity_permutations
+        else:
+            raise NotImplementedError(f"entity_permutations not yet implemented for a general {type(self)}")
+
     def space_dimension(self):
         return self.element.space_dimension()
 
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -48,6 +48,12 @@ class EnrichedElement(FiniteElementBase):
                                        methodcaller("entity_dofs"))
 
     @cached_property
+    def entity_permutations(self):
+        '''Return the map of topological entities to the map of
+        orientations to permutation lists for the finite element'''
+        return concatenate_entity_permutations(self.elements)
+
+    @cached_property
     def _entity_support_dofs(self):
         return concatenate_entity_dofs(self.cell, self.elements,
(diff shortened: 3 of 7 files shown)


wence-:
I kind of would prefer `raise NotImplementedError` over a silent `None`.

ksagiyam:
Yes. I'll fix.

ksagiyam:
Fixed! Firedrake tests pass: https://github.com/firedrakeproject/firedrake/pull/2101

finat/cube.py
```diff
@@ -49,6 +49,13 @@ def _entity_support_dofs(self):
     def entity_dofs(self):
         return self._entity_dofs
 
+    @cached_property
+    def _entity_permutations(self):
+        return flatten_permutations(self.product.entity_permutations())
+
+    def entity_permutations(self):
+        return self._entity_permutations
```
dham, line 57: Do this once and use `functools.cache` if you think it needs caching.

finat/enriched.py
```diff
@@ -166,3 +171,25 @@ def concatenate_entity_dofs(ref_el, elements, method):
             for ent, off in dofs.items():
                 entity_dofs[dim][ent] += list(map(partial(add, offsets[i]), off))
     return entity_dofs
+
+
+def concatenate_entity_permutations(ref_el, elements):
+    """Combine the entity permutations from a list of elements into a combined
+    dict containing the information for the concatenated permutations of all
+    the elements.
```
dham, line 179: Make it clear what gets concatenated.

dham (changes requested):
Just subject to making sure the documentation makes sense to mere mortals.

finat/cube.py
wence-: So I would just `@cached_property` this.

ksagiyam: Fixed.

finat/enriched.py
ksagiyam: Updated doc string.

finat/discontinuous.py
```diff
@@ -36,6 +36,18 @@ def _entity_dofs(self):
...
         return self._entity_dofs
 
+    @cached_property
+    def entity_permutations(self):
+        # Return entity_permutations of the base finite element if it only
+        # has cell degrees of freedom; otherwise entity_permutations is not
+        # yet implemented for DiscontinuousElement.
+        if all(v == []
+               for d in list(sorted(self.element.entity_dofs().keys()))[:-1]
+               for v in self.element.entity_dofs()[d].values()):
```
wence-, line 46: AKA either `self.element.entity_dofs() == self.element.entity_closure_dofs()` or `len(self.space_dimension()) == len(self.element.entity_dofs()[self.cell.get_dimension()][0])`

finat/finiteelementbase.py
```diff
@@ -32,6 +32,31 @@ def entity_dofs(self):
...
+
+        The entity permutations `dict` for the degree 4 Lagrange finite
+        element on the interval, for instance, is given by:
+
+        .. code-block:: python3
+
+            {0: {0: {0: [0]},
+                 1: {0: [0]}},
+             1: {0: {0: [0, 1, 2],
+                     1: [2, 1, 0]}}}
```
wence-, line 50: Are the keys of the innermost dict always just `range(num_orientations)` ? If so, do we need keys?

wence-: If I understand it, this is effectively memoizing the action of each group member for every entity, knowing that the group is a permutation group. Is this right?

ksagiyam: Yes, you are right.
In my current notation, the keys of the inner most dict are indeed always `range(num_orientations)`, and I should have just used `tuple`s instead of `dict`s. If we wanted to adopt PETSc style here, though, would we have something like:
```
{0: {0: {0: [0]},
     1: {0: [0]}},
1: {0: {0: [0, 1, 2],
        -1: [2, 1, 0]}}}
```
as per Table 1.3 of Matt's note?

finat/discontinuous.py
ksagiyam: Now using `self.element.entity_dofs() == self.element.entity_closure_dofs()`. Thanks!

finat/finiteelementbase.py
ksagiyam: As I noted here https://github.com/firedrakeproject/firedrake/pull/2101, using PETSc notation for orientations here would be ideal, but seems not straightforward until we change how we construct `cell_closure`. I wonder if what I'm saying makes sense.

dham (approved):

PR #88, opened 2021-06-07
AWconditioning

FAznaran:
Small changes to the FInAT file for the Arnold--Winther elements, to make the transformation of the internal DOFs consistent with the zany theory, and to post-hoc rescale the DOFs so as to give the mass matrices O(1) conditioning.

### DIFF
--- a/finat/aw.py
+++ b/finat/aw.py
@@ -3,7 +3,9 @@ import numpy
 
 import FIAT
 
-from gem import Literal, ListTensor
+#from gem import Literal, ListTensor
+import gem
+from gem import Literal, ListTensor, Index, Indexed, Product, IndexSum, MathFunction
 
 from finat.fiat_elements import FiatElement
 from finat.physically_mapped import PhysicallyMappedElement, Citations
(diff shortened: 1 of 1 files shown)


finat/aw.py
```diff
@@ -3,7 +3,9 @@
 
 import FIAT
 
-from gem import Literal, ListTensor
+#from gem import Literal, ListTensor
```
rckirby, line 6: Remove the commmented line

finat/aw.py
```diff
@@ -27,23 +29,39 @@ def _edge_transform(T, coordinate_mapping):
     J_np = numpy.array([[J[0, 0], J[0, 1]],
                         [J[1, 0], J[1, 1]]])
     JTJ = J_np.T @ J_np
-
+
```
rckirby, line 32: flake (white space)

finat/aw.py
```diff
@@ -27,23 +29,39 @@ def _edge_transform(T, coordinate_mapping):
     J_np = numpy.array([[J[0, 0], J[0, 1]],
```
rckirby, line 29: Is there syntactic sugar in gem to make this cleaner (so you don't have to explicitly index/wrap)?

rckirby (commented):
I marked a couple of obvious flake things, and you should fix the rest of the ones that are failing.

rckirby (changes requested):
Also, do we want to include a test confirming an O(1) condition number?

wence-:
@FAznaran is it possible to add a test of the conditioning? I suppose the idea is that the mass matrix should have O(1) conditioning on the reference cell? I think one can write a finat-only test to build the mass matrix and then get numeric values out using the gem evaluator.

FAznaran:
What does @pefarrell think? He's said before that if a test depends on Firedrake, it should live there. IIUC [my PR](https://github.com/firedrakeproject/firedrake/pull/2241) is currently failing its own test, I think simply since _this_ PR is not merged, which is an argument for making the test finat-only.

rckirby:
The test I am suggesting could be done entirely in FInAT (+FIAT+numpy)
without loading any Firedrake meshes, solvers, etc.  But they have the
distinct disadvantage of not existing yet.
The Firedrake PR with Firedrake-level tests couldn't be put in as a FInAT
test and would need to stay in Firedrake.

pefarrell:
My opinion is if the (i) FInAT-only test doesn't yet exist and (ii) it's quick to run in the firedrake test suite, then we should just keep things as-is with these PRs.

PR #89, opened 2021-06-18
Dual Evaluation (Round 2)

ReubenHill:
This WIP draft PR gives dual evaluation a second go (in contrast to #57 ) and sets the direction I want to take this in.

For discussion with more work to be done (e.g. I plan to redo the creation of the dual basis of a `FiatElement` starting again from what is currently implemented in tsfc).


finat/tensor_product.py
```diff
@@ -162,6 +162,28 @@ def point_evaluation(self, order, point, entity=None):
...
+            qA, qB = qs
+            Aindices = gem.indices(len(qA.shape))
+            Bindices = gem.indices(len(qB.shape))
+            Q = gem.ComponentTensor(qA[Aindices]*qB[Bindices], Aindices + Bindices)
+            return Q, ps
+        else:
+            raise ValueError(f"{self.__class__.__name__} does not have a dual_basis yet!")
+
+    def dual_evaluation(self, fn):
+        if hasattr(fn, 'factors'):
```
ReubenHill, line 181:
```
if not hasattr(fn, 'factors'):
    return super().dual_evaluation(fn)
```

finat/tensor_product.py
```diff
@@ -162,6 +162,28 @@ def point_evaluation(self, order, point, entity=None):
 
         return self._merge_evaluations(factor_results)
 
+    @property
+    def dual_basis(self):
+        # Outer product the dual bases of the factors
+        qs, pss = zip(*(f.dual_basis for f in self.factors))
+        ps = TensorPointSet(pss)
+        if len(qs) == 2:
```
ReubenHill, line 170: replace with loop over `qs` adding to the expression on each loop

finat/tensorfiniteelement.py
```diff
@@ -120,6 +120,27 @@ def _tensorise(self, scalar_evaluation):
             )
         return result
 
+    @property
+    def dual_basis(self):
+
+        scalar = self._base_element
+        Q, points = scalar.dual_basis
+
+        if len(self._shape) == 2:
```
ReubenHill, line 129: replace with loop over `self._shape` adding to the expression on each loop

.mailmap
```diff
@@ -0,0 +1 @@
+Reuben Nixon-Hill <reuben.nixon-hill10@imperial.ac.uk> <ReubenHill@users.noreply.github.com>
```
wence-, line 1: Is this because you did some "bad" commits in the past?

finat/tensor_product.py
```diff
@@ -162,6 +162,28 @@ def point_evaluation(self, order, point, entity=None):
...
+            # suppose there are two factors
+            qA, qB = qs
+            Aindices = gem.indices(len(qA.shape))
+            Bindices = gem.indices(len(qB.shape))
+            Q = gem.ComponentTensor(qA[Aindices]*qB[Bindices], Aindices + Bindices)
```
wence-, line 175:
```
indices = tuple(gem.indices(len(q.shape)) for q in qs)
Q = gem.ComponentTensor(reduce(operator.mul, (q[i] for q, i in zip(qs, indices)), indices)
```
I suppose.

finat/tensor_product.py
wence-: Reading the description of the interface, I am confused in that this dual_evaluation somehow does `fn(Q)` so can `fn` not return a sum-factorised gem expression?

ReubenHill: I'm afraid I don't quite understand your question. One never evaluates `fn(Q)`, a given function `fn` ought to eat an `AbstractPointSet`

wence-: What I guess I mean is, surely the function should just know if it has factors and do the appropriate thing or not rather than dispatching here?

wence-: But perhaps I just don't understand the way this fits together at the moment.

.mailmap
ReubenHill: Name and email change after I got married - forgot to add it when I last committed to FInAT master.

finat/tensor_product.py
ReubenHill: IIRC the future implementation that @dham and I dreamed up would look something like
```python
if not hasattr(fn, 'factors'): 
    return super().dual_evaluation(fn)
else:
    assert len(fn.factors) == len(self.factors)
    for i, factor in enumerate(self.factors):
        factor_Q, factor_ps = factor.dual_basis
        factor_gem_tensor = factor.dual_evaluation(fn.factors[i])
        ...
        # somehow build up whole dual evaluation using above information
        ...
    return gem_tensor
```
 (caveat: I am not operating at 100% at the moment so I may have recalled wrong)

wence-: OK, that looks fine. I suppose something has to dispatch and this is a reasonable place.

finat/tensorfiniteelement.py
```diff
@@ -120,6 +120,27 @@ def _tensorise(self, scalar_evaluation):
...
+            # Suppose the tensor element has shape (2, 4)
+            # These identity matrices may have difference sizes depending the shapes
+            # tQ = Q ⊗ 𝟙₂ ⊗ 𝟙₄
+            # This can be concretely realised with
+            indices = gem.indices(len(Q.shape))
+            i, j = (gem.Index(extent=self._shape[0]) for _ in range(self._shape[0]))
+            k, l = (gem.Index(extent=self._shape[1]) for _ in range(self._shape[1])) # noqa E741
+            # TODO Need to check how this plays with the transpose argument to TensorFiniteElement.
+            tQ = gem.ComponentTensor(Q[indices]*gem.Delta(i, j)*gem.Delta(k, l), indices + (i, j, k, l))
```
ReubenHill, line 138: There's something wrong here. This gives `tQ` with shape `[Q.shape[0], Q.shape[1], 2, 2, 4, 4]` where I would expect to get `tQ` with shape `[Q.shape[0], Q.shape[1], 2, 4]`

wence-: Why? `CT(Delta(i, j))` has shape `(i.extent, j.extent)`

ReubenHill: I mean that the maths doesn't seem to add up. Ultimately I need to tensor contract this `tQ` with `expr = fn(points)` but it seems that my `tQ` has too much shape.

ReubenHill: I thought that the answer might be
```python
shape_indices = tuple(gem.Index(extent=ex) for ex in Q.shape)
            expansion_indices = tuple(gem.Index(extent=ex) for ex in self._shape)
            tQ = gem.ComponentTensor(Q[shape_indices]*gem.Delta(expansion_indices[0], expansion_indices[1]), shape_indices + expansion_indices )
```
but this seems to cause interpolation to give the wrong answers when compared to projection (in tests that ought to pass)

wence-: I think you want one delta for each shape axis. So something like:
```
is = tuple(gem.Index(extent=d) for d in self._shape)
js = tuple(gem.Index(extent=d) for d in self._shape)

deltas = reduce(gem.Product, (gem.Delta(i, j) for i, j in zip(is, js)))
```

wence-: > I mean that the maths doesn't seem to add up. Ultimately I need to tensor contract this `tQ` with `expr = fn(points)` but it seems that my `tQ` has too much shape.

Can you write it out? Section 2.3 of https://arxiv.org/pdf/1711.02473.pdf may be helpful

wence-: Your `expr` should, I think, have shape `tensor_shape` at each point, is that right? Suppose it is `(2, 3)`. And there are `Q` points, so the total shape is `(Q, 2, 3)`. We index with: `q, i, j` to produce

```
exprqij = expr[q, i, j]
```
Now we contract over `i` and `j` against `tQ` which has indices `q, i, j, k, l` because it looks like
```
tQ = Q[q]*Delta(i, k)*Delta(j, l)
```

So we have:
```
tQexprqkl = IndexSum(tQ*exprqij, (i, j))
```
and if you wrap this up in a CT you get

```
evaluation = CT(tQexprqkl, (q, k, l))
```
which has shape `(Q, 2, 3)` as required?

ReubenHill: Q for a general FIAT element has shape `(num_nodes, num_points, dual_weight_shape[0], ..., weight_shape[n]) ` assuming the `dual_weight` (the `q`s in the relevant code) is not scalar so I think this may be a little more complicated than this

ReubenHill: You also have to deal with `compile_gem` sometimes dropping indices when `expr` is generated, e.g. when your expression is constant

ReubenHill: It's hard this!

wence-: > Q for a general FIAT element has shape `(num_nodes, num_points, dual_weight_shape[0], ..., weight_shape[n]) ` assuming the `dual_weight` (the `q`s in the relevant code) is not scalar so I think this may be a little more complicated than this

I think this doesn't go against what I said. Suppose we make a `TensorFiniteElement(base, shape=(2, 3))`. Where `base` is intrinsically vector-valued (so has shape `(4,)` say). That means that any expression that we interpolate has shape `(4, 2, 3)` and, because the basis function is vector-valued, we want to produce a dof-value that has shape `(2, 3)`.

The `num_nodes` part is "easy" because that's an output axis. So for each node, you evaluate `expr` at some `nQ` points and get back an expression that has shape `(nQ, 4, 2, 3)`. The base element will advertise a dual that has shape (for each node) `(nQ, 4)`, I think. So we have:
```python
expr = ... # shape (nQ, 4, 2, 3)
base_dual = ... # shape `(nQ, 4)`
q = gem.Index()
b, = base.get_value_indices()
i, j = (gem.Index(extent=d) for d in (2, 3)) # generally self._shape
k, l = (gem.Index(extent=d) for d in (2, 3))
deltas = gem.Delta(i, k)*gem.Delta(j, l)
dual = base_dual[q, b]*deltas # no shape, free indices q, b, i, j, k, l
expr = expr[q, b, i, j]
to_contract = dual*expr # no shape, free indices q, b, i, j, k, l
contracted = gem.IndexSum(to_contract, (q, b, i, j)) # no shape, free indices k, l
result = gem.ComponentTensor(contracted, (k, l)) # shape (2, 3), no free indices
```

Is that right?

ReubenHill: > I think you want one delta for each shape axis. So something like:
> 
> ```
> is = tuple(gem.Index(extent=d) for d in self._shape)
> js = tuple(gem.Index(extent=d) for d in self._shape)
> 
> deltas = reduce(gem.Product, (gem.Delta(i, j) for i, j in zip(is, js)))
> ```



> Your `expr` should, I think, have shape `tensor_shape` at each point, is that right? Suppose it is `(2, 3)`. And there are `Q` points, so the total shape is `(Q, 2, 3)`. We index with: `q, i, j` to produce
> 
> ```
> exprqij = expr[q, i, j]
> ```
> 
> Now we contract over `i` and `j` against `tQ` which has indices `q, i, j, k, l` because it looks like
> 
> ```
> tQ = Q[q]*Delta(i, k)*Delta(j, l)
> ```
> 
> So we have:
> 
> ```
> tQexprqkl = IndexSum(tQ*exprqij, (i, j))
> ```
> 
> and if you wrap this up in a CT you get
> 
> ```
> evaluation = CT(tQexprqkl, (q, k, l))
> ```
> 
> which has shape `(Q, 2, 3)` as required?

Okay I think combining these two suggestions is basically right, although you have to be careful to replace the relevant of the `q` indices (which you might have gotten from the shape of `Q`) with with the indices from the point set you contract over (typically the second index though it depends if it's the expression which adds shape or the shape of the points themselves).

finat/finiteelementbase.py
```diff
@@ -144,11 +144,95 @@ def point_evaluation(self, order, refcoords, entity=None):
...
 
+    def dual_basis(self):
+        '''Return a dual evaluation weight tensor Q and point set x to dual
+        evaluate a function fn at.
+
+        The general dual evaluation is then Q * fn(x).
+
+        If the dual weights are scalar then Q, for a general FIAT element, is a
+        matrix with dimensions (num_nodes, num_points) where num_points made a
+        free index that matches the free
```
ReubenHill, line 155: Check this is true for RT and other vector valued fiat elements

ReubenHill: David reckons Q should be rank 2 or 3 for a general FIAT element

finat/finiteelementbase.py
```diff
@@ -144,11 +144,95 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+
+        If the dual weights are tensor valued then Q, for a general FIAT
+        element, is a tensor with dimensions
+        (num_nodes, num_points, dual_weight_shape[0], ..., dual_weight_shape[n])
+        where num_points made a free index that matches the free index of x.
```
ReubenHill, line 160: I should also index out shape since I need to contract that too

ReubenHill: And/or justify why I index out the points (it makes the contraction easier)

ReubenHill: And/or change the interface to not index out the points

finat/finiteelementbase.py
```diff
@@ -144,11 +144,95 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+        basis_indices = tuple(gem.Index(extent=ex) for ex in Q.shape[:1])
+        try:
+            # Try not to bother multiplying by an identity tensor.
+            # NOTE - Since expr can have no free indices at this
+            # point (see TSFC issue #240), there's no easy way to make this
+            # short cut where expr.free_indices == ()
+            assert self.Q_is_identity and expr.free_indices != ()
+            assert len(set(Q.shape)) == 1
+            # replace the basis index with an index of the same extent in
+            # expr.
```
ReubenHill, line 216: move down since this is documenting line 220

finat/fiat_elements.py
```diff
@@ -173,6 +176,120 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                        # have an identity tensor.
+                        self.Q_is_identity = False
+
+        #
+        # CONVERT Q TO gem.Literal (TODO: should be a sparse tensor)
+        #
+
+        # temporary until sparse literals are implemented in GEM which will
+        # automatically convert a dictionary of keys internally.
+        Q = gem.Literal(sparse.as_coo(Q).todense())
```
ReubenHill, line 264: This is a regression which we should do some timing on. Things to test: 3D high order non-tensor product (such as a degree 10 tetrahedron)

ReubenHill: RT or BDM are good things to try

ReubenHill: (test against master! we've introduced a bunch of multiplying by zeros)

finat/tensorfiniteelement.py
```diff
@@ -120,6 +120,57 @@ def _tensorise(self, scalar_evaluation):
...
+        expr_contraction_indices = tuple(gem.Index(extent=d) for d in self._shape)
+        basis_tensor_indices = tuple(gem.Index(extent=d) for d in self._shape)
+        num_factors = 1
+        if hasattr(self.base_element, 'factors'):
+            num_factors = len(self.base_element.factors)
+        elif hasattr(self.base_element, 'product'):
+            num_factors = len(self.base_element.product.factors)
+        # TODO: work out if there are any other cases where the basis indices
+        # in the shape of Q are more than just the first shape index (as in
+        # tensor product elements)
```
ReubenHill, line 167: This might happen with EnrichedElement

finat/fiat_elements.py
ReubenHill: looks to be fine - see for example https://github.com/firedrakeproject/firedrake/pull/2115#issuecomment-888394264

finat/tensor_product.py
```diff
@@ -162,6 +164,64 @@ def point_evaluation(self, order, point, entity=None):
...
+            #
+
+            # NOTE: any shape indices in the expression are because the
+            # expression is tensor valued.
+            # NOTE: here the first num_factors rows of Q are node sets (1 for
+            # each factor)
+            # TODO: work out if there are any other cases where the basis
+            # indices in the shape of the dual basis tensor Q are more than
+            # just the first shape index (e.g. with EnrichedElement)
+            num_factors = total_num_factors(self.factors)
```
wence-, line 202: This feels a little hacky, and seems like it is fragile (I notice that `total_num_factors` has to handle certain elements specially).

Is this interface (returning something with partial shape and partial free indices) the right one for this manipulation and consumers?

I think that at the end of the day we're going to assign something of shape `(num_functionals,)` to an array. But in the TPE case, this is broken down into `(num_functionals_x, num_functionals_y, ...)` ?

wence- (commented):
A first go. I think this mostly looks good and despite the many comments I am probably nitpicking

finat/tensor_product.py
```diff
@@ -162,6 +164,64 @@ def point_evaluation(self, order, point, entity=None):
...
+            assert expr.shape == Q.shape[num_factors:]
+            expr_shape_indices = tuple(gem.Index(extent=ex) for ex in expr.shape)
+            basis_indices = tuple(gem.Index(extent=ex) for ex in Q.shape[:num_factors])
+            # TODO: Work out how to deal with identity Q in this case.
+            dual_evaluation_indexed_sum = gem.optimise.make_product((Q[basis_indices + expr_shape_indices], expr[expr_shape_indices]), x.indices+expr_shape_indices)
+
+            return dual_evaluation_indexed_sum, basis_indices
```
wence-, line 209: I think the generic implementation is:
```
shape_indices = gem.indices(len(expr.shape))
basis_indices = gem.indices(len(Q.shape) - len(expr.shape))
Qf = IndexSum(Q[basis_indices + expr_shape_indices] * expr[expr_shape_indices], x.indices + shape_indices)
Qf = gem.optimise.contraction(Qf)
```

finat/finiteelementbase.py
```diff
@@ -144,11 +144,92 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+
+        # NOTE: any shape indices in the expression are because the expression
+        # is tensor valued.
+        assert expr.shape == Q.shape[1:]
+        expr_shape_indices = tuple(gem.Index(extent=ex) for ex in expr.shape)
+        basis_indices = tuple(gem.Index(extent=ex) for ex in Q.shape[:1])
+        try:
+            assert self.Q_is_identity
+            assert len(set(Q.shape)) == 1
+            assert len(basis_indices) == 1
```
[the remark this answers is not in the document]
ReubenHill: I don't think Q could ever have shape `()` - that would imply 0 nodes in the dual basis

finat/tensor_product.py
ReubenHill: yes it seems you are correct. I'll make all of them follow this

ReubenHill: (i.e. just use this in `FiniteElementBase`)

ReubenHill: I tell a lie it doesn't work for tensorfiniteelement in all cases

wence-: Can you show an example? I have been playing round with this a little.

ReubenHill: Try `tests/regression/test_interpolate_vs_project.py::test_interpolate_vs_project[square-VectorFunctionSpace(BDM2)]` in firedrake: trying to work out what the issue is

ReubenHill: It's either to do with getting the delta elimination to work properly by having the right indices or it's to do with not accounting for `self.base_element.get_value_indices()` in the generic case I think

ReubenHill: I think the issue is that we're contracting over all the expression shape indices which we shouldn't do here - some of them are actually part of the basis

wence-: I think that's right. We're contracting over enough indices, just some of the wrong ones.

wence-: Maybe we have to have the dual_evaluation separate for this case. Perhaps this diff, which I think handles the `transpose` argument correctly?
```
diff --git a/finat/tensorfiniteelement.py b/finat/tensorfiniteelement.py
index 839a009..2db8868 100644
--- a/finat/tensorfiniteelement.py
+++ b/finat/tensorfiniteelement.py
@@ -5,8 +5,6 @@ import numpy
 import gem
 
 from finat.finiteelementbase import FiniteElementBase
-from finat.tensor_product import total_num_factors
-from finat.cube import FlattenedDimensions
 
 
 class TensorFiniteElement(FiniteElementBase):
@@ -124,59 +122,53 @@ class TensorFiniteElement(FiniteElementBase):
 
     @property
     def dual_basis(self):
-
+        if self._transpose:
+            raise NotImplementedError("Not implemented for transposed TFE")
         base = self.base_element
         Q, points = base.dual_basis
 
         # Suppose the tensor element has shape (2, 4)
         # These identity matrices may have difference sizes depending the shapes
         # tQ = Q ⊗ 𝟙₂ ⊗ 𝟙₄
-        Q_shape_indices = gem.indices(len(Q.shape))
-        i_s = tuple(gem.Index(extent=d) for d in self._shape)
-        j_s = tuple(gem.Index(extent=d) for d in self._shape)
-        # we need one delta for each shape axis
-        deltas = reduce(gem.Product, (gem.Delta(i, j) for i, j in zip(i_s, j_s)))
-        # TODO Need to check how this plays with the transpose argument to TensorFiniteElement.
-        tQ = gem.ComponentTensor(Q[Q_shape_indices]*deltas, Q_shape_indices + i_s + j_s)
+        scalar_i = self.base_element.get_indices()
+        scalar_vi = self.base_element.get_value_indices()
+        tensor_i = tuple(gem.Index(extent=d) for d in self._shape)
+        tensor_vi = tuple(gem.Index(extent=d) for d in self._shape)
+        # Couple new basis function and value indices
+        deltas = reduce(gem.Product, (gem.Delta(j, k)
+                                      for j, k in zip(tensor_i, tensor_vi)))
+        if self._transpose:
+            index_ordering = tensor_i + scalar_i + tensor_vi + scalar_vi
+        else:
+            index_ordering = scalar_i + tensor_i + tensor_vi + scalar_vi
 
-        return tQ, points
+        Qi = Q[scalar_i + scalar_vi]
+        Q = gem.ComponentTensor(Qi*deltas, index_ordering)
+        return Q, points
 
     def dual_evaluation(self, fn):
-
         Q, x = self.dual_basis
         expr = fn(x)
-
-        # TODO: Add shortcut (if relevant) for Q being identity tensor
-
+        gem.optimise.contraction(expr)
         # NOTE: any shape indices in the expression are because the expression
         # is tensor valued.
-        assert set(expr.shape) == set(self.value_shape)
-
-        base_value_indices = self.base_element.get_value_indices()
-
-        # Reconstruct the indices from the deltas in dual_basis above and
-        # contract Q with expr
-        Q_shape_indices = tuple(gem.Index(extent=ex) for ex in Q.shape)
-        expr_contraction_indices = tuple(gem.Index(extent=d) for d in self._shape)
-        basis_tensor_indices = tuple(gem.Index(extent=d) for d in self._shape)
-        # NOTE: here the first num_factors rows of Q are node sets (1 for
-        # each factor)
-        # TODO: work out if there are any other cases where the basis
-        # indices in the shape of the dual basis tensor Q are more than
-        # just the first shape index (e.g. with EnrichedElement)
-        if hasattr(self.base_element, 'factors'):
-            num_factors = total_num_factors(self.base_element.factors)
-        elif isinstance(self.base_element, FlattenedDimensions):
-            # Factor might be a FlattenedDimensions which introduces
-            # another factor without having a factors property
-            num_factors = 2
+        assert expr.shape == self.value_shape
+
+        scalar_i = self.base_element.get_indices()
+        scalar_vi = self.base_element.get_value_indices()
+        tensor_i = tuple(gem.Index(extent=d) for d in self._shape)
+        tensor_vi = tuple(gem.Index(extent=d) for d in self._shape)
+
+        if self._transpose:
+            index_ordering = tensor_i + scalar_i + tensor_vi + scalar_vi
         else:
-            num_factors = 1
-        Q_indexed = Q[Q_shape_indices[:num_factors] + base_value_indices + expr_contraction_indices + basis_tensor_indices]
-        expr_indexed = expr[expr_contraction_indices + base_value_indices]
-        dual_evaluation_indexed_sum = gem.optimise.make_product((Q_indexed, expr_indexed), x.indices + base_value_indices + expr_contraction_indices)
-        basis_indices = Q_shape_indices[:num_factors] + basis_tensor_indices
-        return dual_evaluation_indexed_sum, basis_indices
+            index_ordering = scalar_i + tensor_i + tensor_vi + scalar_vi
+
+        Qi = Q[index_ordering]
+        expri = expr[tensor_i + scalar_vi]
+        evaluation = gem.IndexSum(Qi * expri, x.indices + scalar_vi + tensor_i)
+        evaluation = gem.optimise.contraction(evaluation, tensor_vi)
+        return evaluation, scalar_i + tensor_vi
 
     @property
     def mapping(self):
```

wence-: That doesn't work (the old code also not) if the `TensorElement` was made with `symmetry=True` (so that `element.value_shape() != element.reference_value_shape()`

ReubenHill: yes I couldn't work out how to make this work in general. I've added the below as documentation for the tensor finite element case:
```python
    @property
    def dual_basis(self):
        '''Return a dual evaluation gem weight tensor Q and point set x to dual
        evaluate a function fn at.
...
        If the dual basis is of a tensor finite element with some shape 
        (S1, S2, ..., Sn) then the tensor element Q is constructed from the 
        base element's Q (here called q) by taking the outer product with 
        appropriately sized identity matrices:
        Q = q ⊗ 𝟙ₛ₁ ⊗ 𝟙ₛ₂ ⊗ ... ⊗ 𝟙ₛₙ
        '''
```

ReubenHill: > That doesn't work (the old code also not) if the `TensorElement` was made with `symmetry=True` (so that `element.value_shape() != element.reference_value_shape()`

When does that happen? No where in the firedrake test suite covers that

wence-: If I make a `TensorElement("P", triangle, 1, symmetry=True)`.

wence-: But I think the problem is actually not in this code, but rather that in TSFC we need to apply a non-affine mapping to the expression we compile in the case of a symmetries=True TensorElement.

ReubenHill: It looks like FInAT elements don't have a property called reference_value_shape - any ideas on how I'd raise a `NotImplementedError` or `ValueError` in FInAT for this case?

wence-: Can you just do it in Firedrake?
```patch
diff --git a/firedrake/interpolation.py b/firedrake/interpolation.py
index c5db09f6..3ff0974d 100644
--- a/firedrake/interpolation.py
+++ b/firedrake/interpolation.py
@@ -236,6 +236,8 @@ def _interpolator(V, tensor, expr, subset, arguments, access):
     try:
         if not isinstance(expr, firedrake.Expression):
             to_element = create_element(V.ufl_element())
+            if V.ufl_element().mapping() == "symmetries":
+                raise NotImplementedError("Interpolation into tensor spaces with symmetry")
         else:
             # compile_python_kernel code pathway expects base element
             to_element = create_base_element(V.ufl_element())
```

wence-: I think I broadly know what needs to be done, but a bunch of refactoring is required.

ReubenHill: > Maybe we have to have the dual_evaluation separate for this case. Perhaps this diff, which I think handles the `transpose` argument correctly?
> 
> ```diff
> diff --git a/finat/tensorfiniteelement.py b/finat/tensorfiniteelement.py
> index 839a009..9535a6e 100644
> --- a/finat/tensorfiniteelement.py
> +++ b/finat/tensorfiniteelement.py
> @@ -5,8 +5,6 @@ import numpy
>  import gem
>  
>  from finat.finiteelementbase import FiniteElementBase
> -from finat.tensor_product import total_num_factors
> -from finat.cube import FlattenedDimensions
>  
>  
>  class TensorFiniteElement(FiniteElementBase):
> @@ -124,59 +122,51 @@ class TensorFiniteElement(FiniteElementBase):
>  
>      @property
>      def dual_basis(self):
> -
>          base = self.base_element
>          Q, points = base.dual_basis
>  
>          # Suppose the tensor element has shape (2, 4)
>          # These identity matrices may have difference sizes depending the shapes
>          # tQ = Q ⊗ 𝟙₂ ⊗ 𝟙₄
> -        Q_shape_indices = gem.indices(len(Q.shape))
> -        i_s = tuple(gem.Index(extent=d) for d in self._shape)
> -        j_s = tuple(gem.Index(extent=d) for d in self._shape)
> -        # we need one delta for each shape axis
> -        deltas = reduce(gem.Product, (gem.Delta(i, j) for i, j in zip(i_s, j_s)))
> -        # TODO Need to check how this plays with the transpose argument to TensorFiniteElement.
> -        tQ = gem.ComponentTensor(Q[Q_shape_indices]*deltas, Q_shape_indices + i_s + j_s)
> +        scalar_i = self.base_element.get_indices()
> +        scalar_vi = self.base_element.get_value_indices()
> +        tensor_i = tuple(gem.Index(extent=d) for d in self._shape)
> +        tensor_vi = tuple(gem.Index(extent=d) for d in self._shape)
> +        # Couple new basis function and value indices
> +        deltas = reduce(gem.Product, (gem.Delta(j, k)
> +                                      for j, k in zip(tensor_i, tensor_vi)))
> +        if self._transpose:
> +            index_ordering = tensor_i + scalar_i + tensor_vi + scalar_vi
> +        else:
> +            index_ordering = scalar_i + tensor_i + tensor_vi + scalar_vi
>  
> -        return tQ, points
> +        Qi = Q[scalar_i + scalar_vi]
> +        Q = gem.ComponentTensor(Qi*deltas, index_ordering)
> +        return Q, points
>  
>      def dual_evaluation(self, fn):
> -
>          Q, x = self.dual_basis
>          expr = fn(x)
> -
> -        # TODO: Add shortcut (if relevant) for Q being identity tensor
> -
> +        gem.optimise.contraction(expr)
>          # NOTE: any shape indices in the expression are because the expression
>          # is tensor valued.
> -        assert set(expr.shape) == set(self.value_shape)
> -
> -        base_value_indices = self.base_element.get_value_indices()
> -
> -        # Reconstruct the indices from the deltas in dual_basis above and
> -        # contract Q with expr
> -        Q_shape_indices = tuple(gem.Index(extent=ex) for ex in Q.shape)
> -        expr_contraction_indices = tuple(gem.Index(extent=d) for d in self._shape)
> -        basis_tensor_indices = tuple(gem.Index(extent=d) for d in self._shape)
> -        # NOTE: here the first num_factors rows of Q are node sets (1 for
> -        # each factor)
> -        # TODO: work out if there are any other cases where the basis
> -        # indices in the shape of the dual basis tensor Q are more than
> -        # just the first shape index (e.g. with EnrichedElement)
> -        if hasattr(self.base_element, 'factors'):
> -            num_factors = total_num_factors(self.base_element.factors)
> -        elif isinstance(self.base_element, FlattenedDimensions):
> -            # Factor might be a FlattenedDimensions which introduces
> -            # another factor without having a factors property
> -            num_factors = 2
> +        assert expr.shape == self.value_shape
> +
> +        scalar_i = self.base_element.get_indices()
> +        scalar_vi = self.base_element.get_value_indices()
> +        tensor_i = tuple(gem.Index(extent=d) for d in self._shape)
> +        tensor_vi = tuple(gem.Index(extent=d) for d in self._shape)
> +
> +        if self._transpose:
> +            index_ordering = tensor_i + scalar_i + tensor_vi + scalar_vi
>          else:
> -            num_factors = 1
> -        Q_indexed = Q[Q_shape_indices[:num_factors] + base_value_indices + expr_contraction_indices + basis_tensor_indices]
> -        expr_indexed = expr[expr_contraction_indices + base_value_indices]
> -        dual_evaluation_indexed_sum = gem.optimise.make_product((Q_indexed, expr_indexed), x.indices + base_value_indices + expr_contraction_indices)
> -        basis_indices = Q_shape_indices[:num_factors] + basis_tensor_indices
> -        return dual_evaluation_indexed_sum, basis_indices
> +            index_ordering = scalar_i + tensor_i + tensor_vi + scalar_vi
> +
> +        Qi = Q[index_ordering]
> +        expri = expr[tensor_i + scalar_vi]
> +        evaluation = gem.IndexSum(Qi * expri, x.indices + scalar_vi + tensor_i)
> +        evaluation = gem.optimise.contraction(evaluation, tensor_vi)
> +        return evaluation, scalar_i + tensor_vi
>  
>      @property
>      def mapping(self):
> ```

This is great, much neater than what I wrote. What's the reason for `gem.optimise.contraction(expr)`? Also I presume you meant to write `evaluation = gem.optimise.contraction(evaluation)`?

wence-: I'll prepare a PR into this branch...

ReubenHill: FYI I've already added the above - just wanted to annotate it with some comments. I've pushed what I did

finat/finiteelementbase.py
```diff
@@ -144,11 +144,92 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+        # TENSOR CONTRACT Q WITH expr
+        #
+
+        # NOTE: any shape indices in the expression are because the expression
+        # is tensor valued.
+        assert expr.shape == Q.shape[1:]
+        expr_shape_indices = tuple(gem.Index(extent=ex) for ex in expr.shape)
+        basis_indices = tuple(gem.Index(extent=ex) for ex in Q.shape[:1])
+        try:
+            assert self.Q_is_identity
```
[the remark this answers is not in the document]
ReubenHill: `Delta` nodes are better for the `Q_is_identity` case - in that case I wouldn't even both with the sparse literal

finat/fiat_elements.py
```diff
@@ -173,6 +176,119 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+
+                # Ensure all weights have the same shape
+                if last_shape is not None:
+                    assert q_j.shape == last_shape
+                last_shape = q_j.shape
+
+                # Create key into x dict
+                x_key = np.round(x_j.points, x_key_decimals)
+                assert np.allclose(x_j.points, x_key, atol=x_key_atol)
+                x_key = tuple(x_key.flatten())
```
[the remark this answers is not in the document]
ReubenHill: I originally did this because there were cases where the same point was being treated as not-the-same-point to within some small error. This could of course have been because of my initial wrapping in a `PointSet`. I'll test if your re-implementation without rounding still causes any issues (sadly I didn't not the exact test case). I guess even if it does it's probably safer not to force points to be the same within some magical tolerance

finat/finiteelementbase.py
```diff
@@ -144,11 +144,92 @@ def point_evaluation(self, order, refcoords, entity=None):
         :param entity: the cell entity on which to tabulate.
         '''
 
+    @property
+    def dual_basis(self):
+        '''Return a dual evaluation gem weight tensor Q and point set x to dual
+        evaluate a function fn at.
+
+        The general dual evaluation is then Q * fn(x).
```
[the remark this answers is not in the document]
ReubenHill: Yes

finat/finiteelementbase.py
```diff
@@ -144,11 +144,92 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+        If the dual weights are tensor valued then Q, for a general tensor
+        valued FIAT element, is a tensor with dimensions
+        (num_nodes, num_points, dual_weight_shape[0], ..., dual_weight_shape[n])
+        where num_points made a free index that matches the free index of x.
+
+        if the dual basis is of a tensor product or FlattenedDimensions element
+        with N factors then Q in general is a tensor with dimensions
+        (num_nodes_factor1, ..., num_nodes_factorN,
+            num_points_factor1, ..., num_points_factorN,
+            dual_weight_shape[0], ..., dual_weight_shape[n])
```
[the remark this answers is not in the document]
ReubenHill: 👍

finat/finiteelementbase.py
```diff
@@ -144,11 +144,92 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+        '''
+        raise NotImplementedError(
+            f"Dual basis not defined for element {type(self).__name__}"
+        )
+
+    def dual_evaluation(self, fn):
+        '''Return code for performing the dual basis evaluation at the nodes of
+        the reference element. Currently only works for non-derivatives (not
+        implemented) and flat elements (implemented in TensorFiniteElement and
+        TensorProductElement).
```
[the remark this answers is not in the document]
ReubenHill: reworded

finat/tensor_product.py
ReubenHill: I think this was resolved in the discussion at the firedrake meeting on 18/08/2021

finat/quadrature_element.py
```diff
@@ -127,6 +123,19 @@ def basis_evaluation(self, order, ps, entity=None, coordinate_mapping=None):
     def point_evaluation(self, order, refcoords, entity=None):
         raise NotImplementedError("QuadratureElement cannot do point evaluation!")
 
+    @property
+    def dual_basis(self):
+        ps = self._rule.point_set
+        Q = gem.Literal([self._rule.weights])
```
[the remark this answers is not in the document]
ReubenHill: > Do you need to rank-extend the weights when making `Q`, I guess yes, because the first dimension of `Q` is the number of functionals?

Yes, otherwise I could just use `Q = self._rule.weight_expression` I think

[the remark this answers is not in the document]
ReubenHill: > I think that if this a `TensorProductQuadratureRule` then `self._rule.weights` doesn't exist.

Hmm. Possible design mistake somewhere?

finat/quadrature_element.py
```diff
@@ -127,6 +123,19 @@ def basis_evaluation(self, order, ps, entity=None, coordinate_mapping=None):
...
+        # Index out ps.indices from Q
+        assert len(ps.indices) == 1
+        assert Q.shape[1] == ps.indices[0].extent
+        shape_indices = tuple(gem.Index(extent=s) for s in Q.shape)
+        Q = gem.ComponentTensor(
+            gem.Indexed(Q, (shape_indices[0],) + ps.indices + shape_indices[2:]),
+            (shape_indices[0],) + shape_indices[2:])
```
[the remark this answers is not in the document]
ReubenHill: Well based on the code as written you get one functional per weight. So the shape of `Q` is
```
(num_weights, num_points)
```
I know this works at the moment for interpolation onto a `VertexOnlyMesh` (only one quadrature point of weight `1`!) but I'm not quite sure where else one interpolates onto a `QuadratureElement`. 
It is accounted for in the old `c_e_d_e` [here](https://github.com/firedrakeproject/tsfc/blob/b7c66f1ee58637a254ec634ac1d1e369a8c7c41a/tsfc/driver.py#L406-L411):
```python
            # Allow interpolation onto QuadratureElements to refer to the quadrature
            # rule they represent
            if isinstance(to_element, FIAT.QuadratureElement):
                assert allclose(asarray(qpoints), asarray(to_element._points))
                quad_rule = QuadratureRule(point_set, to_element._weights)
                config["quadrature_rule"] = quad_rule
```
so I'd need to know who uses that code path

finat/quadrature_element.py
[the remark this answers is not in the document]
ReubenHill: edit: I mean design mistake by me!

finat/quadrature_element.py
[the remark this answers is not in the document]
ReubenHill: (I've switched to something inspired by your suggestion btw but keeping this unresolved so you can respond)

finat/quadrature_element.py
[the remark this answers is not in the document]
ReubenHill: For now added `NotImplementedError`

finat/fiat_elements.py
```diff
@@ -173,6 +176,119 @@ def point_evaluation(self, order, refcoords, entity=None):
...
+                        # have an identity tensor.
+                        self.Q_is_identity = False
+
+        #
+        # CONVERT Q TO gem.Literal (TODO: should be a sparse tensor)
+        #
+
+        # temporary until sparse literals are implemented in GEM which will
+        # automatically convert a dictionary of keys internally.
+        Q = gem.Literal(sparse.as_coo(Q).todense())
```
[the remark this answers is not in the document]
ReubenHill: Now just need to find a neat way of getting shape...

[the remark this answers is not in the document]
ReubenHill: My suggestion
```python
            ndims = len(list(Q.keys())[0])
            shape = tuple(max(Q.keys(), key=itemgetter(i))[i] for i in range(ndims))
            Qdense = np.zeros(shape)
            for idx, value in Q.items():
                Qdense[idx] = value
            Q = Qdense
```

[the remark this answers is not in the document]
wence-: I think you need to add 1 to all the extents in the shape (your maximum index is say `n` but the shape needs to be `n+1`), since we're zero-indexed, not 1-indexed.

one-liner:

```
shape = tuple(s + 1 for s in map(max, zip(*Q.keys()))
```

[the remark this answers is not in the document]
ReubenHill: 🤦  quite right

One day I will know the language well enough to be able to come up with one liners like that

wence-:
I had a brief go at implementing enriched elements. I think the basic model still works but it plausibly needs more technology in gem/tsfc. In particular, the point set for an enriched dual is the concatenation of the point sets of the subelements. This is reasonably easy to handle by building a new `ConcatPointSet` and using `gem.Concatenate` to represent the sub-structured concatenated points. The same idea doesn't work with the current technology for the Q tensor: all sub-Qs in the enriched element must have the same "value shape", but the point sets have different shape, and they have different numbers of nodes.

`gem.Concatenate` flattens all shapes and adds things up, so `(2, 2) + (6,)` turns into `(10,)` with a way of picking the pieces out. For the `Q` we need to represent concatenation over two axes and reshaping so that we end up with something that looks like `(total_num_duals, total_num_points, *value_shape)`. We can wrap up the duals and points for each sub-element in shape and then concatenate all of that, but then we get `(total_num_duals*total_num_points, *value_shape)`, and there's no way to turn only the shape corresponding to the point set into a free index.

Is it the case that an element will only ever eat a function of its _own_ point set? In which case we can probably square this circle by making the `dual` object something other than directly `gem` and decomposing the pieces in `dual_evaluation` (still need to think about composition of elements). Effectively you would hand code the evaluation of the function on each sub-piece and then `gem.Concatenate` these evaluations together to hand back something that is `(total_num_duals,)` in shape.

I don't think this needs to be fixed for this round, but if anyone has any good ideas I'm all ears.

ReubenHill:
> I had a brief go at implementing enriched elements. I think the basic model still works but it plausibly needs more technology in gem/tsfc. In particular, the point set for an enriched dual is the concatenation of the point sets of the subelements. This is reasonably easy to handle by building a new `ConcatPointSet` and using `gem.Concatenate` to represent the sub-structured concatenated points. The same idea doesn't work with the current technology for the Q tensor: all sub-Qs in the enriched element must have the same "value shape", but the point sets have different shape, and they have different numbers of nodes.
> 
> `gem.Concatenate` flattens all shapes and adds things up, so `(2, 2) + (6,)` turns into `(10,)` with a way of picking the pieces out. For the `Q` we need to represent concatenation over two axes and reshaping so that we end up with something that looks like `(duals, points, *value_shape)`.
> 
> [EDIT: the duals + points don't product together, so we'd need to represent a ragged "3D" array using concatenation which I think we can't do]

@dham and I came to the conclusion that if the two subelements can't be concatenated on the `duals` dimension such that you don't get a ragged tensor then then what you're trying to do is silly and doesn't make much sense.
I don't recall the exact mathematical justification though.

> Is it the case that an element will only ever eat a function of its _own_ point set? In which case we can probably square this circle by making the `dual` object something other than directly `gem` and decomposing the pieces in `dual_evaluation` (still need to think about composition of elements). Effectively you would hand code the evaluation of the function on each sub-piece and then `gem.Concatenate` these evaluations together to hand back something that is `(total_num_duals,)` in shape.

Well you can give an element any function (including one that that returns a constant in all cases) but `dual_evaluation` will, by definition, cause that function to be called on the point set given by that element's `dual_basis`.
As long as the expression you get back from the function is correctly shaped gem then you ought not to get an error.

wence-:
> @dham and I came to the conclusion that if the two subelements can't be concatenated on the `duals` dimension such that you don't get a ragged tensor then then what you're trying to do is silly and doesn't make much sense.
> I don't recall the exact mathematical justification though.

`-ENOPARSE` (too many triple negatives). Is the claim that enriched elements will always concatenate "ragged" duals? I'm not actually sure if that matters for my musings.

wence-:
> Well you can give an element any function (including one that that returns a constant in all cases) but `dual_evaluation` will, by definition, cause that function to be called on the point set given by that element's `dual_basis`.
> As long as the expression you get back from the function is correctly shaped gem then you ought not to get an error.

OK, this actually makes things easier because then in enriched dual evaluation you can pull apart the point set, apply the function to it, contract each subpart with its bit, then concatenate things together at the end.

ReubenHill:
> > @dham and I came to the conclusion that if the two subelements can't be concatenated on the `duals` dimension such that you don't get a ragged tensor then then what you're trying to do is silly and doesn't make much sense.
> > I don't recall the exact mathematical justification though.
> 
> `-ENOPARSE` (too many triple negatives). Is the claim that enriched elements will always concatenate "ragged" duals? I'm not actually sure if that matters for my musings.

Apologies - I mean that if subelement 1 has shape `(dual1, points1, *value_shape1)` and subelement 2 has shape `(dual2, points2, *value_shape2)` then you must have `points1 == points2` and `value_shape1 == value_shape2`

wence-:
> > > @dham and I came to the conclusion that if the two subelements can't be concatenated on the `duals` dimension such that you don't get a ragged tensor then then what you're trying to do is silly and doesn't make much sense.
> > > I don't recall the exact mathematical justification though.
> > 
> > 
> > `-ENOPARSE` (too many triple negatives). Is the claim that enriched elements will always concatenate "ragged" duals? I'm not actually sure if that matters for my musings.
> 
> Apologies - I mean that if subelement 1 has shape `(duals1, points1, *value_shape1)` and subelement 2 has shape `(duals2, points2, *value_shape2)` then you must have `points1 == points2` and `value_shape1 == value_shape2`

I think that I agree that you need `value_shape1 == value_shape2`, however, I don't think you need `points1 == points2`. For example:

```
from ufl import *
from tsfc.finatinterface import create_element
U1 = FiniteElement("P", interval, 1)
U2 = FiniteElement("DP", interval, 1)
V1 = FiniteElement("P", interval, 2)
V2 = FiniteElement("DP", interval, 0)

# an enriched element
Q = TensorProductElement(U1, V2) + TensorProductElement(U2, V1) 

Qe = create_element(Q)

print(Qe.elements[0].dual_basis[1].points.shape)
=> (2, 2)
print(Qe.elements[1].dual_basis[1].points.shape)
=> (6, 2)
```

We use these kind of elements when making spaces on extruded meshes that have different degree in the horizontal and vertical directions.

ReubenHill:
> > > > @dham and I came to the conclusion that if the two subelements can't be concatenated on the `duals` dimension such that you don't get a ragged tensor then then what you're trying to do is silly and doesn't make much sense.
> > > > I don't recall the exact mathematical justification though.
> > > 
> > > 
> > > `-ENOPARSE` (too many triple negatives). Is the claim that enriched elements will always concatenate "ragged" duals? I'm not actually sure if that matters for my musings.
> > 
> > 
> > Apologies - I mean that if subelement 1 has shape `(duals1, points1, *value_shape1)` and subelement 2 has shape `(duals2, points2, *value_shape2)` then you must have `points1 == points2` and `value_shape1 == value_shape2`
> 
> I think that I agree that you need `value_shape1 == value_shape2`, however, I don't think you need `points1 == points2`. For example:
> 
> ```
> from ufl import *
> from tsfc.finatinterface import create_element
> U1 = FiniteElement("P", interval, 1)
> U2 = FiniteElement("DP", interval, 1)
> V1 = FiniteElement("P", interval, 2)
> V2 = FiniteElement("DP", interval, 0)
> 
> # an enriched element
> Q = TensorProductElement(U1, V2) + TensorProductElement(U2, V1) 
> 
> Qe = create_element(Q)
> 
> print(Qe.elements[0].dual_basis[1].points.shape)
> => (2, 2)
> print(Qe.elements[1].dual_basis[1].points.shape)
> => (6, 2)
> ```
> 
> We use these kind of elements when making spaces on extruded meshes that have different degree in the horizontal and vertical directions.

So I think this might work:

Let element 0 have dual basis `Q_0, x_0` and element1 have dual basis `Q_1, x_1` where `Q_0` has shape `(duals0, points0, *value_shape0)` and `Q_1` has shape `(duals1, points1, *value_shape1)`.

We want a new dual set `Q_e, x_e` that we can do a dual evaluation with given some function `f` that acts on `x_e:
```python
expr = f(x_e)
# ... correctly index Q_e and expr
IndexSum(Q_e_indexed * expr_indexed, contraction_indices)
```

So `x_e` is `x_0` and `x_1` concatenated along their first dimension. This can be done by rebuilding from scratch so we get a new free index along `x_e` or by using the concatenation technology you spoke about above.

`Q_e` now needs to be contracted with `f(x_e)` to give the correct dual evaluation expression. I think this just means building a block matrix from `Q_0` and `Q_1` where the two are stacked along the `duals` dimension with `Q_0` above `Q_1`. Any mismatch in the `points` dimension just has to be filled out with zeros to the right. We enforce that `value_shape0 == value_shape1`. `Q_e` now has shape `(duals0+duals1, max(points0, points1), *value_shape0)`

Thoughts?

wence-:
> `Q_e` now needs to be contracted with `f(x_e)` to give the correct dual evaluation expression. I think this just means building a block matrix from `Q_0` and `Q_1` where the two are stacked along the `duals` dimension with `Q_0` above `Q_1`. Any mismatch in the `points` dimension just has to be filled out with zeros to the right. We enforce that `value_shape0 == value_shape1`. `Q_e` now has shape `(duals0+duals1, max(points0, points1), *value_shape0)`
> 
> Thoughts?

Conceptually I think this works, but my concern is how to maintain the structure to pull apart later. In particular, all of the Hdiv/curl elements on quads are built by gluing together pieces with an enriched element. So I want to do the dual evaluation by doing evaluation of the subcomponents (which maintains structure) and then gluing together the result. I can just do this "by hand" in the dual_evaluation routine for enriched elements if I allow myself to break the contract that the `dual_basis` property advertises, but then I can't compose an enriched element with anything, so I'm trying to figure out what the appropriate way of expressing things is that retains the structure.

ReubenHill:
> > `Q_e` now needs to be contracted with `f(x_e)` to give the correct dual evaluation expression. I think this just means building a block matrix from `Q_0` and `Q_1` where the two are stacked along the `duals` dimension with `Q_0` above `Q_1`. Any mismatch in the `points` dimension just has to be filled out with zeros to the right. We enforce that `value_shape0 == value_shape1`. `Q_e` now has shape `(duals0+duals1, max(points0, points1), *value_shape0)`
> > Thoughts?
> 
> Conceptually I think this works, but my concern is how to maintain the structure to pull apart later. In particular, all of the Hdiv/curl elements on quads are built by gluing together pieces with an enriched element. So I want to do the dual evaluation by doing evaluation of the subcomponents (which maintains structure) and then gluing together the result. I can just do this "by hand" in the dual_evaluation routine for enriched elements if I allow myself to break the contract that the `dual_basis` property advertises, but then I can't compose an enriched element with anything, so I'm trying to figure out what the appropriate way of expressing things is that retains the structure.

Well the structure is visible by asking the enriched element for its sub element dual bases.

If you are really keen to hand code a special `dual_evaluation` routine for enriched elements (I presume for some performance boost) then you can have `dual_basis` advertise something of the shape I described and just not use it in `dual_evaluation` in this case (and rather have `dual_evaluation` use the sub element `dual_basis` properties by hand as you describe).
Such a `dual_basis` would still work with the general case `dual_evaluation` routines and would be composable in, e.g., tensor produce elements.
The downside is that you only get any special performance boost when you call the enriched element's `dual_evaluation` routine directly.

Perhaps I'm missing the point though?

wence-:
> Well the structure is visible by asking the enriched element for its sub element dual bases.
> 
> If you are really keen to hand code a special `dual_evaluation` routine for enriched elements (I presume for some performance boost) then you can have `dual_basis` advertise something of the shape I described and just not use it in `dual_evaluation` in this case (and rather have `dual_evaluation` use the sub element `dual_basis` properties by hand as you describe).
> Such a `dual_basis` would still work with the general case `dual_evaluation` routines and would be composable in, e.g., tensor produce elements.
> The downside is that you only get any special performance boost when you call the enriched element's `dual_evaluation` routine directly.
> 
> Perhaps I'm missing the point though?

No, that's exactly it. Up to now, the contract between `dual_basis` and `dual_evaluation` is that they communicate using `gem` and point sets. This is composable, since if I'm an element modifier, I don't need to know details of the element I am modifying to correctly construct my dual basis. If we break this contract here (to provide structured evaluation for the specific case of enrichedelement), then element modifiers _do_ need to know the details.

ReubenHill:
> > Well the structure is visible by asking the enriched element for its sub element dual bases.
> > If you are really keen to hand code a special `dual_evaluation` routine for enriched elements (I presume for some performance boost) then you can have `dual_basis` advertise something of the shape I described and just not use it in `dual_evaluation` in this case (and rather have `dual_evaluation` use the sub element `dual_basis` properties by hand as you describe).
> > Such a `dual_basis` would still work with the general case `dual_evaluation` routines and would be composable in, e.g., tensor produce elements.
> > The downside is that you only get any special performance boost when you call the enriched element's `dual_evaluation` routine directly.
> > Perhaps I'm missing the point though?
> 
> No, that's exactly it. Up to now, the contract between `dual_basis` and `dual_evaluation` is that they communicate using `gem` and point sets. This is composable, since if I'm an element modifier, I don't need to know details of the element I am modifying to correctly construct my dual basis. If we break this contract here (to provide structured evaluation for the specific case of enrichedelement), then element modifiers _do_ need to know the details.

I what way are we breaking any contract? As far as I have gathered so far the special enrichedelement `dual_evaluation` routine is just a performance boost but you could just as well use the standard routine on my suggested enrichedelement `dual_basis`

wence-:
> I what way are we breaking any contract? As far as I have gathered so far the special enrichedelement `dual_evaluation` routine is just a performance boost but you could just as well use the standard routine on my suggested enrichedelement `dual_basis`

Ideally, I don't want a situation where two almost identical elements have vastly difference performance properties for interpolation. That would be the case if we go down a "special case enriched element dual evaluation" I _think_. For example, if one takes a tensor product of enriched elements, then the dual_evaluation of the tensor product element won't see any structure. In 3D this is a `p**6` vs `p**4` algorithm which can be painful (especially if everything else is `p**4`).

ReubenHill:
> > I what way are we breaking any contract? As far as I have gathered so far the special enrichedelement `dual_evaluation` routine is just a performance boost but you could just as well use the standard routine on my suggested enrichedelement `dual_basis`
> 
> Ideally, I don't want a situation where two almost identical elements have vastly difference performance properties for interpolation. That would be the case if we go down a "special case enriched element dual evaluation" I _think_. For example, if one takes a tensor product of enriched elements, then the dual_evaluation of the tensor product element won't see any structure. In 3D this is a `p**6` vs `p**4` algorithm which can be painful (especially if everything else is `p**4`).

You might have already said as much, but how does the special enriched element `dual_evaluation` routine (i.e. pulling apart the sub elements, doing the dual evaluation, then putting them back together)  improve performance? All I can think of right now is avoiding multiplying by zeroes - something that goes away when we have sparse tensors.

wence-:
> You might have already said as much, but how does the special enriched element `dual_evaluation` routine (i.e. pulling apart the sub elements, doing the dual evaluation, then putting them back together) improve performance? All I can think of right now is avoiding multiplying by zeroes - something that goes away when we have sparse tensors.

If the subelements have structure that can be exploited, that structure is maintained (e.g. consider enriched for two tensor product elements). This is more than just not multiplying by zeros since it now admits a computationally lower complexity algorithm.

ReubenHill:
> > You might have already said as much, but how does the special enriched element `dual_evaluation` routine (i.e. pulling apart the sub elements, doing the dual evaluation, then putting them back together) improve performance? All I can think of right now is avoiding multiplying by zeroes - something that goes away when we have sparse tensors.
> 
> If the subelements have structure that can be exploited, that structure is maintained (e.g. consider enriched for two tensor product elements). This is more than just not multiplying by zeros since it now admits a computationally lower complexity algorithm.

Ah and I presume my proposed solution, without new `gem` nodes for concatenating matrices as I suggest, would cause structure to be lost?

wence-:
> Ah and I presume my proposed solution, without new `gem` nodes for concatenating matrices as I suggest, would cause structure to be lost?

Yes, because once you splat to `Literal`s you can't recover it.

ReubenHill:
> > Ah and I presume my proposed solution, without new `gem` nodes for concatenating matrices as I suggest, would cause structure to be lost?
> 
> Yes, because once you splat to `Literal`s you can't recover it.

Okay so we need something more clever that allows us to recover such information. Perhaps it could be built into `SparseLiteral`?

PR #91, opened 2021-07-13
Compare to spatial dim when making QuadratureElement

### LINKED ISSUE #90: QuadratureElement is not comparing the cell and quadrature dimensions correctly
Suppose I want to define a FunctionSpace with a Quadrature element on a hexahedral mesh:
```python
from firedrake import *
mesh = ExtrudedMesh(UnitSquareMesh(1,1,quadrilateral=True),layers=1)
Qe = FiniteElement("Quadrature", mesh.ufl_cell(), degree=3, quad_scheme="default")
Q = FunctionSpace(mesh, Qe)
```
The snippet above gives the following error:
```
~/firedrake-dev-20200825-mpich/src/FInAT/finat/quadrature_element.py in __init__(self, fiat_ref_cell, rule)
     44             raise TypeError("rule is not an AbstractQuadratureRule")
     45         if fiat_ref_cell.get_dimension() != rule.point_set.dimension:
---> 46             raise ValueError("Cell dimension does not match rule's point set dimension")
     47         self._rule = rule
     48 
ValueError: Cell dimension does not match rule's point set dimension
```

This is because the cell dimension comes out as a tuple:
```
(Pdb++) fiat_ref_cell.get_dimension()
(2, 1)
(Pdb++) rule.point_set.dimension
3
```

ReubenHill:
This fixes #90: the wrong FIAT reference element dimension was being
checked against in the case of a tensor product cells

### DIFF
--- a/finat/quadrature_element.py
+++ b/finat/quadrature_element.py
@@ -42,7 +42,7 @@ class QuadratureElement(FiniteElementBase):
         self.cell = fiat_ref_cell
         if not isinstance(rule, AbstractQuadratureRule):
             raise TypeError("rule is not an AbstractQuadratureRule")
-        if fiat_ref_cell.get_dimension() != rule.point_set.dimension:
+        if fiat_ref_cell.get_spatial_dimension() != rule.point_set.dimension:
             raise ValueError("Cell dimension does not match rule's point set dimension")
         self._rule = rule
 


ReubenHill:
Can be merged once https://github.com/firedrakeproject/firedrake/pull/2147 passes

wence- (approved):
Thanks, hit go if we get the pass.

PR #92, opened 2021-08-18
Ensure literals in point evaluation are floats

wence-:


finat/fiat_elements.py
```diff
@@ -275,7 +275,8 @@ def is_const(expr):
             assert table.shape[-1] == m
             zerocols = np.isclose(abs(table).max(axis=tuple(range(table.ndim - 1))), 0.0)
             if all(np.logical_or(const_mask, zerocols)):
-                vals = base_values_sympy[const_mask]
+                # Casting is safe by assertion of is_const
```
ReubenHill, line 278: I don't see such an assertion anywhere. Did you mean to add it and forget?

ReubenHill (commented):
Colour me amazed if this fixes https://github.com/firedrakeproject/firedrake/issues/2095 . How on earth did you work it out?

finat/fiat_elements.py
wence-: The `is_const` function does it. It tries to cast the contents to `float` and sets `const_mask` to True in those cases, otherwise it sets them to `float`

ReubenHill: got it

ReubenHill (approved):

wence-:
> Colour me amazed if this fixes [firedrakeproject/firedrake#2095](https://github.com/firedrakeproject/firedrake/issues/2095) . How on earth did you work it out?

I knew that this was coming from runtime point evaluation, so I tried to see where the complex float was appearing, and the ciarlet point evaluation was the only candidate.

PR #93, opened 2021-08-19
Some updates

wence-:


finat/finiteelementbase.py
```diff
@@ -194,48 +194,30 @@ def dual_evaluation(self, fn):
...
-            Qfn = gem.IndexSum(Q[basis_indices + expr_shape_indices] * expr[expr_shape_indices], x.indices + expr_shape_indices)
-            Qfn = gem.optimise.contraction(Qfn)
-
-        return Qfn, basis_indices
+        Qi = Q[basis_indices + shape_indices]
+        expri = expr[shape_indices]
+        evaluation = gem.IndexSum(Qi * expri, x.indices + shape_indices)
+        # Don't want to factorise over the shape indices in the
+        # contraction, so exclude those from the optimisation routines.
+        evaluation = gem.optimise.contraction(evaluation, shape_indices)
```
wence-, line 213: Note, this is a bit of a sledgehammer and not perfectly targeted. What we should do is implement something similar to the `spectral` optimisation mode that tsfc applies when compiling forms. arguably this is easier because we know this contraction is just a big linear thing already....

finat/fiat_elements.py
```diff
@@ -190,103 +190,50 @@ def dual_basis(self):
...
+            pts = tuple(sorted(point_dict.keys()))
+            kstart, kend = seen[pts]
+            for p, k in zip(pts, range(kstart, kend)):
+                for weight, cmp in point_dict[p]:
+                    Q[(i, k, *cmp)] = weight
+
         x = PointSet(allpts)
+        if all(len(set(key)) == 1 and np.isclose(weight, 1) and len(key) == 2
+               for key, weight in Q.items()):
+            # Identity, is this the most general case?
```
ReubenHill, line 223: No. You are looking purely for an identity matrix whereas my existing check would find any identity-like tensor I that acted on a tensor T to produce T.

ReubenHill: The `all(len(set(key)) == 1` is fine but `np.isclose(weight, 1) and len(key) == 2` are too restrictive. 

I think we get an identity tensor `Q` of shape `(X, X, X)` (say) indexed as `Q[i, j, k]` whenever we have 
```python
np.isclose(Q[i, j, k], 1) and i==j==k
```

ReubenHill: I'm not sure yet what that looks like in terms of kronecker deltas

ReubenHill: I guess it's probably something like an outer product of a bunch of Deltas but I'm not sure (and google isn't being helpful). This is why i just flagged this case in the other code and didn't do the contraction when I identified identity

wence-: So I want to understand exactly what you mean by "Identity" here. Which will probably help me.

I think that you have a pointset (with shape `(p,)`), and evaluate the function (on an expression which has shape `(q1, q2)` say) to get something with shape `(p, q1, q2)`. Now you have a Q tensor with shape `(n, p, q1, q2)` and you want that:

```
CT(sum(Q[i, j, k, l]*expr[j, k, l], (j, k, l)), (i,)) == expr
```
?

But this seems not right because the left hand side has shape `(n,)` and the right hand side has shape `(p, q1, q2)`

wence-: I suppose this only works at least when `p == n` (which is the case if all the nodes are point evaluations?)

ReubenHill: it's the same optimisation that already existed in the TSFC dual eval code for point evaluations, yes

wence-: What about the other part of the question?

ReubenHill: such that rather than doing any contraction you have that 
```
evaluation_expr, basis_indices = element.dual_evaluation(fn)
```
has `basis_indices = x.indices` and `evaluation_expr = fn(x)` (where `x` is from `element.dual_basis`)

wence-: I think this means that `evaluation_expr` must have shape `(p,) == (x.indices[0].extent,)` (with my example above), and so there will never be `q1` and `q2`?

ReubenHill: > What about the other part of the question?

By `Q` being identity I mean that `Q` doesn't actually do anything numerically to `expr = fn(x)` when contracted along `x.indices + expr_shape_indices`: once the code has run it would yield the same thing as `expr`

wence-: I think that if I contract over the `expr_shape_indices` then my output expression must be of lower rank than `expr` (unless `expr_shape_indices == ()`), what am I missing?

ReubenHill: > I think this means that `evaluation_expr` must have shape `(p,) == (x.indices[0].extent,)` (with my example above), and so there will never be `q1` and `q2`?

Yes I think you are correct given your shape argument.

ReubenHill: I think there might be an equivalent optimisation in there for tensor point sets where you don't have expression shape to contract over but I'm not quite sure what it is yet and I certainly hadn't implemented it up to this point

finat/tensor_product.py
```diff
@@ -166,15 +165,11 @@ def point_evaluation(self, order, point, entity=None):
...
-        self.Q_is_identity = all(f.Q_is_identity for f in self.factors)
+        qs, pss = zip(*(factor.dual_basis for factor in self.factors))
         ps = TensorPointSet(pss)
-        tuples_of_q_shape_indices = tuple(gem.indices(len(q.shape)) for q in qs)
-        q_muls = reduce(operator.mul, (q[i] for q, i in zip(qs, tuples_of_q_shape_indices)))
-        # Flatten to get final shape indices
-        q_shape_indices = tuple(index for tuple_of_indices in tuples_of_q_shape_indices for index in tuple_of_indices)
-        Q = gem.ComponentTensor(q_muls, q_shape_indices)
+        qis = tuple(q[gem.indices(len(q.shape))] for q in qs)
+        indices = tuple(chain(*(q.index_ordering() for q in qis)))
```
ReubenHill, line 171: Ooh, I must remember about `Indexed.index_ordering()`

ReubenHill (approved):

finat/fiat_elements.py
wence-: I believe that with the implementation via deltas we have here, this "just works", at least, the code comes out looking right.

This is the kernel on a quad interpolate Q1 into Q2 if I replace the `Delta` with just a tabulated `Q` for the factors:
```
#include <stdint.h>

void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[3 * 3] = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0 };
  double const t1[3 * 2] = { 1.0, 0.0, 0.5, 0.5, 0.0, 1.0 };
  double t2[3 * 2];
  double t3[2];
  double t4[2];

  for (int32_t p1 = 0; p1 <= 2; ++p1)
  {
    for (int32_t i = 0; i <= 1; ++i)
      t2[2 * p1 + i] = 0.0;
    for (int32_t i_0 = 0; i_0 <= 2; ++i_0)
      for (int32_t i_1 = 0; i_1 <= 1; ++i_1)
        t2[2 * p1 + i_1] = t2[2 * p1 + i_1] + t0[3 * p1 + i_0] * t1[2 * i_0 + i_1];
  }
  for (int32_t p0 = 0; p0 <= 2; ++p0)
  {
    for (int32_t i_2 = 0; i_2 <= 1; ++i_2)
    {
      t3[i_2] = 0.0;
      for (int32_t i_3 = 0; i_3 <= 2; ++i_3)
        t3[i_2] = t3[i_2] + t0[3 * p0 + i_3] * t1[i_2 + 2 * i_3];
    }
    for (int32_t i_4 = 0; i_4 <= 1; ++i_4)
    {
      t4[i_4] = 0.0;
      for (int32_t i_5 = 0; i_5 <= 1; ++i_5)
        t4[i_4] = t4[i_4] + w_0[i_4 + 2 * i_5] * t3[i_5];
    }
    for (int32_t p1_0 = 0; p1_0 <= 2; ++p1_0)
      for (int32_t i_6 = 0; i_6 <= 1; ++i_6)
        A[3 * p0 + p1_0] = A[3 * p0 + p1_0] + t2[2 * p1_0 + i_6] * t4[i_6];
  }
}
```

And this is the kernel with the deltas left in:

```
void expression_kernel(double *__restrict__ A, double const *__restrict__ w_0)
{
  double const t0[3 * 2] = { 1.0, 0.0, 0.5, 0.5, 0.0, 1.0 };
  double t1[2];

  for (int32_t p0 = 0; p0 <= 2; ++p0)
  {
    for (int32_t i = 0; i <= 1; ++i)
    {
      t1[i] = 0.0;
      for (int32_t i_0 = 0; i_0 <= 1; ++i_0)
        t1[i] = t1[i] + t0[2 * p0 + i_0] * w_0[i + 2 * i_0];
    }
    for (int32_t p1 = 0; p1 <= 2; ++p1)
      for (int32_t i_1 = 0; i_1 <= 1; ++i_1)
        A[3 * p0 + p1] = A[3 * p0 + p1] + t0[2 * p1 + i_1] * t1[i_1];
  }
}
```

(The former has a flop count of 132, the latter 60)

wence-: Note how all the multiplications by the identity matrix have disappeared.

ReubenHill: Excellent. I have also notices a flop count decrease on the CG1->RT2 script I tested in the firedrake PR (looks like better loop nesting) and general performance improvements all round

ReubenHill: How do we test if sum factorisation is actually working?

wence-: If we merge firedrakeproject/tsfc#256 (or cherry pick https://github.com/firedrakeproject/tsfc/pull/256/commits/496b32460d7c796e2decb41f4a703abf730b196f from that branch, since the complex version of the tests is still broken so the merge is blocked), then when the kernel you get back from the `c_e_d_e` call has a `flop_count` property. On tensor product cells for Q and DQ spaces, these flop counts should asymptotically go like `p^{d+1}` (when interpolating from (say) Q_{p-1} to Q_{p}). That's for simple interpolation of a coefficient, you can do the same thing for interpolation of more complicated expressions.

ReubenHill: And what's a nice everyday example of a tensor product element someone might be using?

wence-: `FiniteElement("Q", hexahedron, p)`

PR #98, opened 2021-09-03
Dual for hdiv/hcurl element wrappers

wence-:
Progress towards having duals for all finat element modifiers. The big missing one now is `EnrichedElement`.

This change also showed up a bug in the implementation of the dual for tensor product elements (if any of the factors had value_shape).

### DIFF
--- a/finat/hdivcurl.py
+++ b/finat/hdivcurl.py
@@ -74,6 +74,19 @@ class WrapperElementBase(FiniteElementBase):
         core_eval = self.wrappee.point_evaluation(order, refcoords, entity)
         return self._transform_evaluation(core_eval)
 
+    @property
+    def dual_basis(self):
+        Q, x = self.wrappee.dual_basis
+        beta = self.get_indices()
+        zeta = self.get_value_indices()
+        # Index out the basis indices from wrapee's Q, to get
+        # something of wrappee.value_shape, then promote to new shape
+        # with the same transform as done for basis evaluation
+        Q = gem.ListTensor(self.transform(gem.partial_indexed(Q, beta)))
+        # Finally wrap up Q in shape again (now with some extra
+        # value_shape indices)
+        return gem.ComponentTensor(Q[zeta], beta + zeta), x
+
 
 class HDivElement(WrapperElementBase):
     """H(div) wrapper element for tensor product elements."""
--- a/finat/tensor_product.py
+++ b/finat/tensor_product.py
@@ -171,9 +171,16 @@ class TensorProductElement(FiniteElementBase):
         # Outer product the dual bases of the factors
         qs, pss = zip(*(factor.dual_basis for factor in self.factors))
         ps = TensorPointSet(pss)
-        qis = tuple(q[gem.indices(len(q.shape))] for q in qs)
-        indices = tuple(chain(*(q.index_ordering() for q in qis)))
-        Q = gem.ComponentTensor(reduce(gem.Product, qis), indices)
+        # Naming as _merge_evaluations above
+        alphas = [factor.get_indices() for factor in self.factors]
+        zetas = [factor.get_value_indices() for factor in self.factors]
+        # Index the factors by so that we can reshape into index-shape
+        # followed by value-shape
+        qis = [q[alpha + zeta] for q, alpha, zeta in zip(qs, alphas, zetas)]
+        Q = gem.ComponentTensor(
+            reduce(gem.Product, qis),
+            tuple(chain(*(alphas + zetas)))
+        )
         return Q, ps
 
     @cached_property


ReubenHill:
> This change also showed up a bug in the implementation of the dual for tensor product elements (if any of the factors had value_shape).

I was pretty sure that there are places where this was tested for in either `firedrake/tests/regression/test_interpolation_nodes.py`, `firedrake/tests/regression/test_interpolate_vs_project.py`or `firedrake/tests/vertexonly/`

Is this branch secretly sitting on top of changes to TSFC and Firedrake?

ReubenHill:
Ah, I take it this will fix the `VectorElement(RaviartThomas, ...)` example?

wence-:
> Ah, I take it this will fix the VectorElement(RaviartThomas, ...) example?

No, it fixes `TensorProductElement(RaviartThomas, Lagrange)` (say).

wence-:
The commit message of 3f378b3a578af652ff029369ab52f53ab0fcb93f explains what's going on.

ReubenHill:
> > Ah, I take it this will fix the VectorElement(RaviartThomas, ...) example?
> 
> No, it fixes `TensorProductElement(RaviartThomas, Lagrange)` (say).

🤦  Right, of course. I'm still surprised this didn't get caught. I guess HDiv/HCurl are the only places in the test suite this gets tested for. 

Well I'm pleased to see this working so straightforwardly. Would you mind briefly explaining the maths of what's going on (ideally as a comment)? Also can you please open a firedrake CI branch with the `xfail`s removed so we can see this in action?

ReubenHill:
> The commit message of [3f378b3](https://github.com/FInAT/FInAT/commit/3f378b3a578af652ff029369ab52f53ab0fcb93f) explains what's going on.

So it seems you're somehow changing the definition of Q to include `index_shape` as well as `value_shape`. Is this because the contraction doesn't work properly and, if so, why? Is the expression advertising the wrong shape?

wence-:
> So it seems you're somehow changing the definition of Q to include `index_shape` as well as `value_shape`. Is this because the contraction doesn't work properly and, if so, why? Is the expression advertising the wrong shape?

`Q` for a raviart-thomas 1 element in 2D has (after making the point a free index) shape `(3, 2)`. If we tensor this with a P0 element on an interval (shape `(1, )`), then we get naively get something of shape `(3, 2, 1)`. But the functional indices need to come before the shape indices, so we need to spit out something of shape `(3, 1, 2)` which we do by separately indexing the sub-qs by indices for their functionals (the index-shape) and then their values (the value-shape) and then wrapping the product in a component tensor that first hands out index-shape and then value-shape.

wence-:
> Would you mind briefly explaining the maths of what's going on (ideally as a comment)?

The hdiv and hcurl element modifiers in finat take a wrappee element and promote the basis_evaluation of the wrappee to the correct vector shape (see `select_hdiv_transformer` and `select_hcurl_transformer` in hdivcurl.py). So that's what I'm doing here for the `Q`, I get the wrappee `Q` with shape `(nfunctionals, wrappee_value_shape)` and promote it to the wrapped value_shape.

finat/tensor_product.py
```diff
@@ -167,9 +167,15 @@ def dual_basis(self):
...
-        Q = gem.ComponentTensor(reduce(gem.Product, qis), indices)
+        alphas = [factor.get_indices() for factor in self.factors]
+        zetas = [factor.get_value_indices() for factor in self.factors]
+        # Index the factors by so that we can reshape into index-shape
+        # followed by value-shape
+        qis = [q[alpha + zeta] for q, alpha, zeta in zip(qs, alphas, zetas)]
+        Q = gem.ComponentTensor(
+            reduce(gem.Product, qis),
+            tuple(chain(*(alphas + zetas)))
+        )
```
ReubenHill, line 178: Looks good to me, though the naming of `alphas` suggests it refers to a geometric shape when really its just the indices over all the bases. Perhaps rename? (Or tell me I've misunderstood!)

finat/hdivcurl.py
```diff
@@ -70,6 +70,14 @@ def point_evaluation(self, order, refcoords, entity=None):
         core_eval = self.wrappee.point_evaluation(order, refcoords, entity)
         return self._transform_evaluation(core_eval)
 
+    @property
+    def dual_basis(self):
+        Q, x = self.wrappee.dual_basis
+        beta = self.get_indices()
+        zeta = self.get_value_indices()
+        Q = gem.ListTensor(self.transform(gem.partial_indexed(Q, beta)))
+        return gem.ComponentTensor(Q[zeta], beta + zeta), x
```
ReubenHill, line 79: This needs some explanatory comments (along the lines of what you have written in [this comment](https://github.com/FInAT/FInAT/pull/98#issuecomment-913717573).

I'm also surprised to see a `gem.ListTensor` and `gem.partial_indexed` appear here. What's going on on that line?

See also the other comment about beta.

ReubenHill (commented):
This looks correct as far as I've managed to understand. It's nice, in particular, to see that `transform` 'just works' though I'd like to see some more comments about the dance you're doing in `dual_basis`

finat/tensor_product.py
wence-: I'm just mimicking the naming in `_merge_evaluations`.

finat/hdivcurl.py
wence-: `Q` has shape `(*self.index_shape, *wrappee.value_shape)`, `transform` promotes something of `wrappee.value_shape` to a `list` of `self.value_shape` which is then `ListTensor`ed. So now we have `Q` with shape `self.value_shape` (good) and free indices `beta`, which we need to turn into a componenttensor.

ReubenHill: Okay so tell me if this is correct:

```
gem.partial_indexed(Q, beta))
```
makes sure your wrappe element's `Q` only has shape `wrappee.value_shape` which is then transformed with `self.transform`.

After being `self.transform`ed we now have a `self.value_shape` long list of tensors which need to be stacked (i.e. wrapped in a `gem.ListTensor`) to get a new `Q`.
This new `Q` has all the right content for our new element (though I'm not sure what that content is in this case).
 - `Q.shape` is `zeta` (?)
 - `Q` has any number of free indices

Lastly we transform our `Q` to match the API by indexing out the shape then re-adding it such that we keep x.indices as our free indices.

finat/tensor_product.py
ReubenHill: In which case please say as much or copy across the explanatory comments from `_merge_evaluations`

finat/hdivcurl.py
wence-: > Okay so tell me if this is correct:
> 
> ```
> gem.partial_indexed(Q, beta))
> ```
> 
> makes sure your wrappe element's `Q` only has shape `wrappee.value_shape` which is then transformed with `self.transform`.
> 
> After being `self.transform`ed we now have a `self.value_shape` long list of tensors which need to be stacked (i.e. wrapped in a `gem.ListTensor`) to get a new `Q`.
> This new `Q` has all the right content for our new element (though I'm not sure what that content is in this case).

Think of an example where we're hdiving a tensor product of P x DP on a quadrilateral. Each of the wrappee element functionals are scalar point evaluations, the hdiv functionals are just these scalar evaluations promoted to be vectors dotted with an appropriate normal. Since we're on the reference cell and know what the normal is, this just turns the functional into a vector `[0, -v]` (evaluate the second component of the vector multiply by minus-one).

The scalar functional might just be something like point evaluation, but the transformed functional is point evaluation of `-1 * third component`, so the initial functional is turned into `[0, 0, -v]`

> 
> * `Q.shape` is `zeta` (i.e. self.value_shape) (?)

Yes.

> * `Q` has any number of free indices though we know that they include `beta` (the indices along the basis)

By design, it must have beta + x.indices, I think.

> Lastly we transform our `Q` to match the API by indexing out the shape then re-adding it such that we keep `x.indices` as our free indices.

Yes.

ReubenHill: Okay good. If you can add some comments here saying (briefly) what you've put here then I think this can be resolved

wence-: Added a short comment.

finat/tensor_product.py
wence-: Done.

ReubenHill (approved):

ReubenHill:
Before merging we will need to get rid of the xfails on the firedrake tests

wence-:
> Before merging we will need to get rid of the xfails on the firedrake tests

Those still fail because they need enriched as well....

ReubenHill:
> > Before merging we will need to get rid of the xfails on the firedrake tests
> 
> Those still fail because they need enriched as well....

Right so we need a pure HDiv/HCurl test. Is that possible with pure firedrake?

ReubenHill:
(i.e. what's an element that would do that?)

wence-:
Make an extruded mesh of an interval and then make an `HDiv(TensorProductElement(FiniteElement("P", interval, 1), FiniteElement("DP", interval, 0)))` or similar (that is the "horizontal" part of an RT element)

PR #104, opened 2022-11-14
orientation: handle extrinsic orientations in tensor product elements

ksagiyam:
Enhance `entity_permutations` of tensorproduct elements so that "extrinsic" orientation can happen when mapping the reference cell to the physical cell.

Example (Interval x Interval x Interval):

Extrinsic orientations: X -> y, Y->z, Z->x, etc. (3! ways)

Intrinsic orientations: X->-X, Y->-Y, etc. (2^3 ways per extrinsic orientation)

This gives 3! * (2^3) = 48 ways of mapping the reference cell to the physical cell.

### DIFF
--- a/finat/tensor_product.py
+++ b/finat/tensor_product.py
@@ -1,4 +1,5 @@
 from functools import reduce
+import itertools
 from itertools import chain, product
 from operator import methodcaller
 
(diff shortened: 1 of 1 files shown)


finat/tensor_product.py
```diff
@@ -1,5 +1,5 @@
 from functools import reduce
-from itertools import chain, product
+from itertools import chain, product, permutations as itertools_permutations
```
wence-, line 2: Better to just `import itertools` and then use `itertools.permutations` below.

ksagiyam: Thanks. Fixed!

dham (approved):
subject to the firedrake PR landing

PR #105, opened 2023-01-11
Expand EnrichedElement

pbrubeck:
Expand nested `EnrichedElement`s into a single `EnrichedElement`.

This fixes the issue https://github.com/firedrakeproject/firedrake/issues/2713

### DIFF
--- a/finat/enriched.py
+++ b/finat/enriched.py
@@ -16,7 +16,7 @@ class EnrichedElement(FiniteElementBase):
     basis functions of several other finite elements."""
 
     def __new__(cls, elements):
-        elements = tuple(sum([e.elements if isinstance(e, EnrichedElement) else )e for e in elements], tuple()))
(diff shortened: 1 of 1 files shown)


finat/enriched.py
```diff
@@ -16,7 +16,7 @@ class EnrichedElement(FiniteElementBase):
     basis functions of several other finite elements."""
 
     def __new__(cls, elements):
-        elements = tuple(elements)
+        elements = tuple(sum([e.elements if isinstance(e, EnrichedElement) else (e,) for e in elements], tuple()))
```
wence-, line 19:
```
from itertools import chain
...
elements = tuple(chain.from_iterable(e.elements if isinstance(e, EnrichedElement) else (e,) for e in elements))
```

wence-:
Can you rebase/merge? Looks like we were still testing against 3.6 (I have updated to 3.8 on trunk)

PR #115, opened 2023-11-21
Tabulate Ciarlet generically

### LINKED ISSUE #114: CI failing due to FIAT changes
Two FInAT tests are [currently failing in CI](https://github.com/FInAT/FInAT/actions/runs/6933464251/job/18859666971?pr=113). I have tracked this down to a change somewhere in FIAT made in the last 2 weeks. Specifically I have validated that the tests pass using commit `f628a01efe9bcdc511ce8a91176756845d30add2` but do not with the latest `master`.

pbrubeck:
It seems that we did not really need a special case for point evaluation of Ciarlet elements. 
Instead of directly tabulating an element on a sympy Variable, we were evaluating the expansion set and manipulating the expansion coefficients trying to simplify tabulations that result in constants, e. g. second derivatives of Lagrange(2). Operating at the expansion coefficient level will only simplify constants when the constant function is a member of the expansion set, which is not true for a LagrangeLineExpansionSet.

It turns out that a special case for Ciarlet is not necessary, as `gem.ListTensor` has constant folding.
https://github.com/firedrakeproject/tsfc/blob/5515b2ae3c19e41bf80f347e7259fe9f5d86a9a2/gem/gem.py#L724

Tests that were failing are passing with this PR and https://github.com/firedrakeproject/fiat/pull/54

fixes #114 while keeping the LagrangeLineExpansionSet

### DIFF
--- a/finat/fiat_elements.py
+++ b/finat/fiat_elements.py
@@ -3,7 +3,6 @@ import sympy as sp
 from functools import singledispatch
 
 import FIAT
-from FIAT.polynomial_set import mis, form_matrix_product
 
 import gem
 from gem.utils import cached_property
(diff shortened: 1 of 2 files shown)


finat/fiat_elements.py
```diff
@@ -267,13 +265,7 @@ def mapping(self):
             return result
 
 
-@singledispatch
 def point_evaluation(fiat_element, order, refcoords, entity):
```
wence-, line 268: You could just push this implementation into the body of the `point_evaluation` method now if you think it would be worth it.

wence-: The best kind

requirements-git.txt
```diff
@@ -1,3 +1,3 @@
-git+https://github.com/firedrakeproject/fiat.git#egg=fenics-fiat
+git+https://github.com/firedrakeproject/fiat.git@pbrubeck/simplify-lagrange
```
dham, line 1:
```suggestion
git+https://github.com/firedrakeproject/fiat.git#egg=fenics-fiat
```

dham (approved):

PR #116, opened 2023-11-22
Allow embedded_subdegree and embedded_superdegree to handle tuples

connorjward:
This appears to fix https://github.com/firedrakeproject/firedrake/issues/3251. However, I'm not familiar enough with UFL to be confident that this is the right thing to do. Is it legal for a "base" finite element to have non-integer `degree`?

### DIFF
--- a/finat/ufl/finiteelement.py
+++ b/finat/ufl/finiteelement.py
@@ -234,3 +234,19 @@ class FiniteElement(FiniteElementBase):
                 None,
                 self.quadrature_scheme(),
                 self.variant())
+
+    @property
+    def embedded_subdegree(self):
+        """Return embedded subdegree."""
+        if isinstance(self.degree(), int):
+            return self.degree()
+        else:
+            return min(self.degree())
+
+    @property
+    def embedded_superdegree(self):
+        """Return embedded superdegree."""
+        if isinstance(self.degree(), int):
+            return self.degree()
+        else:
+            return max(self.degree())


connorjward:
@mscroggs sorry to keep bothering you about this. Do you think that what I've done here makes sense? Specifically the issue we have is the current solution breaks for the following:
```py
Vtrace = FunctionSpace(mesh, "HDiv Trace", degree=(1, 1))
```

connorjward:
@mscroggs sorry to keep bothering you about this. Do you think that what I've done here makes sense? Specifically the issue we have is the current solution breaks for the following:
```py
Vtrace = FunctionSpace(mesh, "HDiv Trace", degree=(1, 1))
```

mscroggs:
The base element in UFL (https://github.com/FEniCS/ufl/blob/3eda4f9dcc6d2d04fd9d4b8e58576b45d60b34fa/ufl/finiteelement.py#L31) does not have a `degree`: it only needs the embedded super- and sub-degrees.

So the degree function is a finat.ufl function, and so it can do anything you like.

mscroggs (approved):

connorjward:
> The base element in UFL (https://github.com/FEniCS/ufl/blob/3eda4f9dcc6d2d04fd9d4b8e58576b45d60b34fa/ufl/finiteelement.py#L31) does not have a `degree`: it only needs the embedded super- and sub-degrees.
> 
> So the degree function is a finat.ufl function, and so it can do anything you like.

Got it. Thanks!

PR #123, opened 2024-04-04
C0/C1 Macroelements

pbrubeck:
Advertise a reference complex in case the basis functions are defined on a split cell.

Implement the HCT triangle and its basis transformation.

Depends on https://github.com/firedrakeproject/fiat/pull/64

### DIFF
--- a/finat/bell.py
+++ b/finat/bell.py
@@ -11,7 +11,7 @@ from finat.physically_mapped import PhysicallyMappedElement, Citations
 class Bell(PhysicallyMappedElement, ScalarFiatElement):
     def __init__(self, cell, degree):
         if degree != 5:
-            raise ValueError("Degree must be 5 for Bell element")
+            raise ValueError("Degree must be 3 for Bell element")
         if Citations is not None:
             Citations().register("Bell1969")
         super().__init__(FIAT.Bell(cell))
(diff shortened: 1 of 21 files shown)


finat/bell.py
```diff
@@ -26,63 +26,63 @@ def basis_transformation(self, coordinate_mapping):
...
-                for i in range(21):
-                    V[i, 6*v+1+k] = V[i, 6*v+1+k] / h[v]
+                V[s+1+k, :] /= h[v]
             for k in range(3):
-                for i in range(21):
-                    V[i, 6*v+3+k] = V[i, 6*v+3+k] / (h[v]*h[v])
+                V[s+3+k, :] /= (h[v]*h[v])
 
-        return ListTensor(V.T)
+        return ListTensor(V)
```
wence-, line 85: This is unrelated to macroelements and just constructing `V.T` directly?

finat/cell_tools.py
```diff
@@ -0,0 +1,8 @@
+import FIAT
+
+
+def max_complex(complexes):
+    """Find the maximal complex in a list of cell complexes.
+    This is a pass-through from FIAT so that FInAT clients
+    (e.g. tsfc) don't have to directly import FIAT."""
+    return FIAT.reference_element.max_complex(complexes)
```
wence-, line 8: This could just be:
```
from FIAT.reference_element import max_complex
```
Maybe with a comment. Then whatever docstring FIAT's `max_complex` has is available.

finat/hct.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self, cell, degree):
+        if degree != 3:
+            raise ValueError("Degree must be 3 for HCT element")
+        if Citations is not None:
+            Citations().register("Clough1965")
+        super().__init__(FIAT.HsiehCloughTocher(cell))
+
+    def basis_transformation(self, coordinate_mapping):
+        # Jacobians at edge midpoints
+        J = coordinate_mapping.jacobian_at([1/3, 1/3])
```
wence-, line 21: That's definitely the jacobian at the cell centre.

finat/hct.py
```diff
@@ -0,0 +1,68 @@
...
+                for j in range(d):
+                    V[s+1+j, s+1+i] = J[j, i]
+
+        for e in range(3):
+            s = 3 * voffset + e
+            v0id, v1id = [i * voffset for i in range(3) if i != e]
+
+            # nhat . J^{-T} . t
+            foo = (rns[e, 0]*(J[0, 0]*pts[e, 0] + J[1, 0]*pts[e, 1])
+                   + rns[e, 1]*(J[0, 1]*pts[e, 0] + J[1, 1]*pts[e, 1]))
```
wence-, line 50: I think this is $\hat{n} \cdot J^T t$, not $\hat{n} \cdot J^{-T} t$.

Could be written as: 

`gem.partial_indexed(rns, (e, )) @ (J.T @ gem.partial_indexed(pts, (e,)))`

I think.

finat/hct.py
pbrubeck: Also the Jacobian everywhere on the cell if it is a constant. What is the right thng here, Hermite does not assume constant Jacobian, how well does the transformation theory extend to non-constant Jacobians?

finat/bell.py
pbrubeck: Correct, I took Bell as template for HCT but I also cleaned up Bell

finat/hct.py
wence-: I recall having some discussions with @rckirby but we never came to a conclusion (nor did we write any maths down). Maybe it's "trivial" and as long as the scaling objects are picked at the appropriate place and the refinement of the mesh converges to affine mapped then everything falls out in the wash.

rckirby: Section 3.4.1 of the original zany element paper says it works, at least for Hermite.  I haven't ever tried this at all, not sure what the delicacies are.

finat/bell.py
rckirby: Bell is not the right go-by.  It was implemented based on Argyris, which used the point normal for edge dof.  It's actually just on a branch (need to fix this!) but Wu-Xu used moments of normal derivatives, which turn out much cleaner.  The Wu-Xu element has all the dofs we need and some extra ones (second normal derivatives on edges), so this is the right go-by for us (see recent paper by Bock/Cotter/Kirby in CAMWA for a write-up).  However, that plus some aspects from Bell would get us reduced HCT.

dham (approved):

PR #124, opened 2024-04-19
Create ufl_signature property

Ig-dolci:
This PR addresses an issue with `ufl` warnings that can be disruptive for users. The warnings arise when users define code like:

```python
quad_rule = finat.quadrature.make_quadrature(V.finat_element.cell, V.ufl_element().degree(), "KMV")
time_term = (1 / (c * c)) * (u - 2.0 * u_n + u_nm1) / Constant(dt**2) * v * dx(scheme=quad_rule)
```

resulting in the following warnings:

```
/Users/ddolci/tes_fire_install/firedrake_test/src/ufl/ufl/utils/sorting.py:86: UserWarning: Applying str() to a metadata value of type QuadratureRule, don't know if this is safe.
  warnings.warn(f"Applying str() to a metadata value of type {type(value).__name__}, "
/Users/ddolci/tes_fire_install/firedrake_test/src/ufl/ufl/utils/sorting.py:86: UserWarning: Applying str() to a metadata value of type QuadratureRule, don't know if this is safe.
  warnings.warn(f"Applying str() to a metadata value of type {type(value).__name__}, 
```

This PR introduces a `Quadrature.ufl_signature` property to resolve these warnings.

### DIFF
--- a/finat/quadrature.py
+++ b/finat/quadrature.py
@@ -89,6 +89,10 @@ class QuadratureRule(AbstractQuadratureRule):
     @cached_property
     def weight_expression(self):
         return gem.Indexed(gem.Literal(self.weights), self.point_set.indices)
+    
+    @property
+    def ufl_signature(self):
+        return type(self).__name__ + str(self._parameters)
 
 
 class TensorProductQuadratureRule(AbstractQuadratureRule):
--- a/finat/quadrature.py
+++ b/finat/quadrature.py
@@ -76,9 +76,9 @@ class QuadratureRule(AbstractQuadratureRule):
     """Generic quadrature rule with no internal structure."""
 
     def __init__(self, point_set, weights):
-        weights = numpy.asarray(weights)
-        assert len(point_set.points) == len(weights)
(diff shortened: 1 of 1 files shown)


finat/quadrature.py
```diff
@@ -90,6 +90,10 @@ def point_set(self):
     def weight_expression(self):
         return gem.Indexed(gem.Literal(self.weights), self.point_set.indices)
 
+    @property
+    def ufl_signature(self):
+        return type(self).__name__ + str(self._parameters)
+
```
dham, line 96:
```suggestion
```

finat/quadrature.py
```diff
@@ -90,6 +90,9 @@ def point_set(self):
     def weight_expression(self):
         return gem.Indexed(gem.Literal(self.weights), self.point_set.indices)
 
+    @property
```
wence-, line 93: *suggestion*: Use a `cached_property` here, since `str`ifying the weights is potentially expensive.

PR #128, opened 2024-06-05
GEM: sympy2gem conditional support

pbrubeck:
This is required to make MG work with macro elements

### DIFF
--- a/finat/sympy2gem.py
+++ b/finat/sympy2gem.py
@@ -1,5 +1,6 @@
 from functools import singledispatch, reduce
 
+import numpy
 import sympy
 try:
     import symengine
--- a/finat/sympy2gem.py
+++ b/finat/sympy2gem.py
@@ -65,3 +65,72 @@ def sympy2gem_symbol(node, self):
 @sympy2gem.register(symengine.Rational)
 def sympy2gem_rational(node, self):
     return gem.Division(*(map(self, node.as_numer_denom())))
+
+
+@sympy2gem.register(sympy.Abs)
+@sympy2gem.register(symengine.Abs)
+def sympy2gem_abs(node, self):
+    return gem.MathFunction("abs", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Not)
+@sympy2gem.register(symengine.Not)
+def sympy2gem_not(node, self):
+    return gem.LogicalNot(*map(self, node.args))
+
+
+@sympy2gem.register(sympy.Or)
+@sympy2gem.register(symengine.Or)
+def sympy2gem_or(node, self):
+    return reduce(gem.LogicalOr, map(self, node.args))
+
+
+@sympy2gem.register(sympy.And)
+@sympy2gem.register(symengine.And)
+def sympy2gem_and(node, self):
+    return reduce(gem.LogicalAnd, map(self, node.args))
+
+
+@sympy2gem.register(sympy.Eq)
+@sympy2gem.register(symengine.Eq)
+def sympy2gem_eq(node, self):
+    return gem.Comparison("==", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Gt)
+def sympy2gem_gt(node, self):
+    return gem.Comparison(">", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Ge)
+def sympy2gem_ge(node, self):
+    return gem.Comparison(">=", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Lt)
+@sympy2gem.register(symengine.Lt)
+def sympy2gem_lt(node, self):
+    return gem.Comparison("<", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Le)
+@sympy2gem.register(symengine.Le)
+def sympy2gem_le(node, self):
+    return gem.Comparison("<=", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Piecewise)
+@sympy2gem.register(symengine.Piecewise)
+def sympy2gem_conditional(node, self):
+    expr = None
+    pieces = []
+    for v, c in node.args:
+        if c == True:  # noqa: E712
+            expr = self(v)
+            break
+        pieces.append((v, c))
+    if expr is None:
+        expr = gem.Literal(float("nan"))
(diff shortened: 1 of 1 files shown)


finat/sympy2gem.py
```diff
@@ -65,3 +65,48 @@ def sympy2gem_symbol(node, self):
 @sympy2gem.register(symengine.Rational)
 def sympy2gem_rational(node, self):
     return gem.Division(*(map(self, node.as_numer_denom())))
+
+
+@sympy2gem.register(sympy.Abs)
+@sympy2gem.register(symengine.Abs)
+def sympy2gem_abs(node, self):
+    return gem.MaxValue(*map(self, node.args), gem.Literal(0))
```
wence-, line 73: Surely not?!

pbrubeck: There's no `gem.Abs`, or I am wrong?

finat/sympy2gem.py
```diff
@@ -65,3 +65,48 @@ def sympy2gem_symbol(node, self):
...
+def sympy2gem_conditional(node, self):
+    # TODO nested conditionals for more than two conditional pairs
+    assert len(node.args) == 2
+    for val, cond in node.args:
+        if cond == True:  # noqa: E712
+            otherwise = self(val)
+        else:
+            value = self(val)
+            condition = self(cond)
+    return gem.Conditional(condition, value, otherwise)
```
wence-, line 112: How about something like this:

```python
has_true = False
for (v, c) in node.args:
    pieces.append((v, c))
    if c == True:
        has_true = True
        break
(v, c), *pieces = reversed(pieces)
expr = self(v) if has_true else gem.Conditional(self(c), self(v), gem.Literal(float("nan")))

for (v, c) in pieces:
    expr = gem.Conditional(self(c), self(v), expr)
return expr
```

finat/sympy2gem.py
wence-: Use `gem.MathFunction("abs", *map(self, node.args))`, but more concretely `max(-1, 0) != abs(-1)` !

finat/sympy2gem.py
```diff
@@ -65,3 +65,51 @@ def sympy2gem_symbol(node, self):
...
+
+
+@sympy2gem.register(sympy.Piecewise)
+@sympy2gem.register(symengine.Piecewise)
+def sympy2gem_conditional(node, self):
+    expr = None
+    pieces = []
+    for v, c in node.args:
+        if c == True:  # noqa: E712
+            expr = self(v)
```
wence-, line 108:
```suggestion
           expr = self(v)
           break
```

So that you don't generate more expression than you need to (this also matches the documented semantics of sympy.Piecewise

finat/sympy2gem.py
```diff
@@ -65,3 +65,51 @@ def sympy2gem_symbol(node, self):
...
+            break
+        pieces.append((v, c))
+    if expr is None:
+        expr = gem.Literal(float("nan"))
```
wence-, line 112: You could push this into the initialisation of `expr = None` since it will be unilaterally overwritten only if the `c == True` condition is hit.

wence- (dismissed):

finat/sympy2gem.py
```diff
@@ -65,3 +65,72 @@ def sympy2gem_symbol(node, self):
...
+    return reduce(gem.LogicalAnd, map(self, node.args))
+
+
+@sympy2gem.register(sympy.Eq)
+@sympy2gem.register(symengine.Eq)
+def sympy2gem_eq(node, self):
+    return gem.Comparison("==", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Gt)
```
pbrubeck, line 100: `symengine.Gt` and `symengine.Ge`  are not the usual wrapper classes, but `cyfunction`s instead.

finat/sympy2gem.py
```diff
@@ -65,3 +65,72 @@ def sympy2gem_symbol(node, self):
...
+    expr = None
+    pieces = []
+    for v, c in node.args:
+        if c == True:  # noqa: E712
+            expr = self(v)
+            break
+        pieces.append((v, c))
+    if expr is None:
+        expr = gem.Literal(float("nan"))
+    for v, c in reversed(pieces):
```
pbrubeck, line 134: @wence- are you sure about reversing the order here?

wence-: Pretty sure, you want the last condition to be made "first", because you keep wrapping the new conditions outside.

Consider:

```
(c1, v1), (c2, v2), (c3, v3), (True, v4)
```

I think that wants to be turned into:

```
Conditional(c1, v1, Conditional(c2, v2, Conditional(c3, v3, v4)))
```

So if you go from the back to the front, you build up the expressions:

```
expr = v4
expr = Conditional(c3, v3, expr) # v4
expr = Conditional(c2, v2, expr) # 
expr = Conditional(c1, v1, expr)
```
Which is what you want, no?

finat/sympy2gem.py
```diff
@@ -65,3 +65,72 @@ def sympy2gem_symbol(node, self):
...
+    return gem.Comparison("<=", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Piecewise)
+@sympy2gem.register(symengine.Piecewise)
+def sympy2gem_conditional(node, self):
+    expr = None
+    pieces = []
+    for v, c in node.args:
+        if c == True:  # noqa: E712
```
dham, line 128:
```suggestion
        if isinstance(c, (bool, numpy.bool) and c :
```

dham (approved):

finat/sympy2gem.py
```diff
@@ -65,3 +66,72 @@ def sympy2gem_symbol(node, self):
...
+    return gem.Comparison("<=", *map(self, node.args))
+
+
+@sympy2gem.register(sympy.Piecewise)
+@sympy2gem.register(symengine.Piecewise)
+def sympy2gem_conditional(node, self):
+    expr = None
+    pieces = []
+    for v, c in node.args:
+        if isinstance(c, (bool, numpy.bool)) and c:
```
pbrubeck, line 129: @dham we also need to consider `sympy.logic.boolalg.BooleanTrue`, should we revert to `c == True`?

PR #129, opened 2024-06-26
High order HCT

pbrubeck:
Add basis transformation for high-order HCT


finat/argyris.py
```diff
@@ -1,42 +1,83 @@
...
             Citations().register("Argyris1968")
-        super().__init__(FIAT.QuinticArgyris(cell))
+        if variant is None:
+            variant = "integral"
+        if variant == "integral":
+            fiat_element = FIAT.Argyris(cell, degree, variant=variant)
+        elif variant == "point":
+            if degree != 5:
+                raise ValueError("Degree must be 5 for 'point' variant of Argyris")
+            fiat_element = FIAT.QuinticArgyris(cell)
```
rckirby, line 58: I thought the point variant with degree 5 in FIAT would be used.  I think we should just have `Argyris`.  
Also, shouldn't we have `NotImplemented` instead -- I think the FIAT element is fine, we just haven't done the transformation theory for point values.

finat/argyris.py
```diff
@@ -48,47 +89,53 @@ def basis_transformation(self, coordinate_mapping):
...
-            v0id, v1id = [i for i in range(3) if i != e]
-            for i in range(21):
-                V[i, 18+e] = 2*V[i, 18+e] / (h[v0id] + h[v1id])
+        for v in sorted(top[0]):
+            for k in range(sd):
+                V[:, voffset*v+1+k] *= 1 / h[v]
+            for k in range((sd+1)*sd//2):
+                V[:, voffset*v+3+k] *= 1 / (h[v]*h[v])
+
+        eoffset = 2 * q - 1
```
rckirby, line 135: The exact nature of the edge dof scaling here is going to depend on the variant of integral vs point values.  We should get that straightened out (tests of mass matrix conditioning will reveal this).

test/test_argyris.py
```diff
@@ -0,0 +1,74 @@
...
+    # i = len(edofs[2][0])
+    # offset = len(edofs[1][0]) + i
+    # print()
+    # print(Mh.T[-offset:-i])
+    # print(M.T[-offset:-i])
+    assert np.allclose(M, Mh, atol=1E-9)
+
+    assert np.allclose(finat_vals, phys_vals[:numbfs])
+
+
```
rckirby, line 67: We should be able to test mass conditioning on a single triangle of size h since the scale separation resulting from the wrongly -scaled dofs is a function of size and not being connected to many other cells (like, say, the condition number for Poisson).

test/test_hct.py
```diff
@@ -23,18 +23,19 @@ def phys_cell(request):
...
     ref_complex = element.get_reference_complex()
-    sd = ref_complex.get_spatial_dimension()
     top = ref_complex.get_topology()
     pts = []
-    for cell in top[sd]:
-        pts.extend(ref_complex.make_points(sd, cell, degree+sd+1))
+    for dim in top:
+        for entity in top[dim]:
+            pts.extend(ref_complex.make_points(dim, entity, degree, variant="gll"))
     return pts
```
rckirby, line 33: Again, a harness for testing mass matrix conditioning on one cell (as a function of h) would be good.  You can do this in a separate file and cover Hermite/Argyris/HCT/Morley all in one test.

rckirby (changes requested):
It's hard to approve until we have Firedrake-level tests confirming we're doing the right thing with this branch, but so far so good.  I'm asking for a test of mass matrix conditioning (one cell is sufficient) and clarifying the role of `QuinticArgyirs` versus `Argyris` with point variant and degree=5.

finat/argyris.py
```diff
@@ -1,42 +1,78 @@
...
 import FIAT
 
-from gem import Literal, ListTensor
+from gem import Literal, ListTensor, partial_indexed
 
 from finat.fiat_elements import ScalarFiatElement
 from finat.physically_mapped import PhysicallyMappedElement, Citations
 
 
+def _edge_transform(V, voffset, fiat_cell, moment_deg, coordinate_mapping, avg=False):
```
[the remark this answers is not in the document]
rckirby: Since `_edge_transform` is used in both Argyris & HCT elements, does it make sense to lift that out into a separate file?

rckirby (approved):
One suggestion for refactoring, but looks fine as long as the tests pass.

PR #131, opened 2024-07-12
HCT: basis transformation for hierarchical edge functions

pbrubeck:
Automatically infer the Jacobi weights in the basis transformation to match the order of the jet at the vertices, this choice ensures a hierarchical basis.


dham (commented):
Yes, subject to passing Firedrake tests.

rckirby (approved):

PR #134, opened 2024-08-12
Use md5 to hash FiniteElementBase

JDBetteridge:
The current implementation produces a different hash for different Python invocations. This means the value cannot safely be used for a (persistent) disk cache.

The new implementation replaces the inbuilt Python `hash` with `md5`, which produces the same result across invocations.

Someone with better knowledge of FInAT should double check that all relevant data is encoded in `self._ufl_hash_data_()` , which I believe is just the repr of whatever the derived object is.

### DIFF
--- a/finat/ufl/finiteelementbase.py
+++ b/finat/ufl/finiteelementbase.py
@@ -85,7 +85,7 @@ class FiniteElementBase(AbstractFiniteElement):
 
     def __hash__(self):
         """Compute hash code for insertion in hashmaps."""
-        return int.from_bytes(md5(self._ufl_hash_data_().encode()).digest())
+        return int.from_bytes(md5(self._ufl_hash_data_().encode()).digest(), byteorder='big')
 
     def __eq__(self, other):
         """Compute element equality for insertion in hashmaps."""
--- a/finat/ufl/finiteelementbase.py
+++ b/finat/ufl/finiteelementbase.py
@@ -12,6 +12,7 @@
 # Modified by Matthew Scroggs, 2023
 
 from abc import abstractmethod, abstractproperty
+from hashlib import md5
 
 from ufl import pullback
 from ufl.cell import AbstractCell, as_cell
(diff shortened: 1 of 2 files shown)


rckirby (approved):
I'm not up on how UFL hashing works (or md5), but if we're triggering recompilation and this fixes that issue while keeping Firedrake passing tests, I'm happy to approve.

finat/ufl/finiteelementbase.py
```diff
@@ -84,7 +85,7 @@ def _ufl_signature_data_(self):
 
     def __hash__(self):
         """Compute hash code for insertion in hashmaps."""
-        return hash(self._ufl_hash_data_())
+        return int.from_bytes(md5(self._ufl_hash_data_().encode()).digest())
```
wence-, line 88: TODO: check if the different "variant"s hash differently

JDBetteridge: dham made the same suggestion and I will double check/write a test. I think it should be fine since:

1. `self._ufl_hash_data_()` in `finiteelementbase.py` is
```python
def _ufl_hash_data_(self):
        """Doc."""
        return repr(self)
```

2. `repr(self)` of `FiniteElement` in `finiteelement.py` is
```python
def __repr__(self):
        """Format as string for evaluation as Python object."""
        return self._repr
```

3. And, finally, in `__init__` of `FiniteElement` in `finiteelement.py`
```python
self._repr = "FiniteElement(%s, %s, %s%s%s)" % (
        repr(self.family()), repr(self.cell), repr(self.degree()), quad_str, var_str)
```
where `var_str` contains the variant.

I'll come back to this once I've actually finished fixing the parallel caching behaviour in [PyOP2](https://github.com/OP2/PyOP2/pull/724)/[Firedrake](https://github.com/firedrakeproject/firedrake/pull/3730).

JDBetteridge: Test added. I've only added a test for a couple of elements, but this should be sufficient to check the failure mode we saw previously.

I think this solution is sufficient when the finite element (Python object) is a subclass instance of `finat.ufl.finiteelementbase.FiniteElementBase` as these objects have a good repr. However, it isn't sufficient if the object is a subclass instance of `finat.finiteelementbase.FiniteElementBase` since the repr isn't implemented, contains a memory address and is very unsafe!

I'm trying to think of an instance where we might need to create a cache key based on the latter, otherwise I think this is sufficient. (@connorjward @dham)

connorjward: I think that until we meet a case where we need safe hashing for FInAT elements we should not spend time fixing them. Sometime in the near future @dham is going to do a refactor of this bit of the code so making changes now might be unproductive anyway.

PR #139, opened 2024-09-14
Add physically mapped Stokes elements

pbrubeck:
Generalize Piola transformation theory for newly added Stokes elements.

Depends on https://github.com/firedrakeproject/fiat/pull/84.


finat/arnold_qin.py
```diff
@@ -0,0 +1,70 @@
...
+            rows = numpy.asarray(transform(fiat_cell, J, detJ, f))
+            fdofs = dofs[sd-1][f]
+            bfs = edofs[sd-1][f][1:]
+            V[numpy.ix_(bfs, fdofs)] = rows[..., :len(fdofs)]
+        return ListTensor(V.T)
+
+
+class ReducedArnoldQin(ArnoldQin):
+    def __init__(self, cell, degree=None):
+        super(ReducedArnoldQin, self).__init__(cell, degree)
```
wence-, line 54:
```suggestion
        super().__init__(cell, degree)
```

finat/physically_mapped.py
```diff
@@ -168,6 +169,39 @@
...
+  author={Arnold, Douglas N and Qin, Jinshui},
+  journal={Advances in computer methods for partial differential equations},
+  volume={7},
+  pages={28--34},
+  year={1992}
+}
+""")
+    Citations().add("BernardiRaugel1985", """
+@article{bernardi-raugel-0,
+    AUTHOR = {Bernardi, Christine and Raugel, Geneviv\\`eve},
```
wence-, line 197:
```suggestion
    AUTHOR = {Bernardi, Christine and Raugel, Genevi\\`eve},
```
One too many vs

finat/physically_mapped.py
```diff
@@ -306,3 +340,37 @@ def physical_vertices(self):
...
+        detA = A[0, 0] * determinant(A[1:, 1:])
+        cols = numpy.ones(A.shape[1], dtype=bool)
+        for j in range(1, n):
+            cols[j] = False
+            detA += (-1)**j * A[0, j] * determinant(A[1:][:, cols])
+            cols[j] = True
+        return detA
+
+
+def adjugate(A):
```
wence-, line 363: FWIW, I think we should introduce determinant and related primitives in gem which TSFC can be lower to numerically stable evaluations, especially since people keep on wanting to do 4D stuff.

finat/piola_mapped.py
```diff
@@ -0,0 +1,84 @@
+import numpy
+from gem import Literal
+
+
+def determinant(A):
```
rckirby, line 5: This is about the worst possible algorithm for `n x n` determinants.  I assume the usage is for a small and probably symbolic matrix?  If so, should this (and the `adjugate` function be lifted into GEM someplace?

finat/piola_mapped.py
```diff
@@ -0,0 +1,84 @@
...
+        rows[i] = False
+        for j in range(A.shape[1]):
+            cols[j] = False
+            C[j, i] = (-1)**(i+j)*determinant(A[rows, :][:, cols])
+            cols[j] = True
+        rows[i] = True
+    return C
+
+
+def piola_inverse(fiat_cell, J, detJ):
```
rckirby, line 40: The inverse of something is adjugate over determinant, so the inverse of the  Piola map is just the adjugate of the Jacobian since the determinants cancel?

test/test_zany_mapping.py
```diff
@@ -100,3 +105,104 @@ def test_argyris_point(ref_cell, phys_cell):
     finat_element = finat.Argyris(ref_cell, variant="point")
     phys_element = type(finat_element.fiat_equivalent)(phys_cell, variant="point")
     check_zany_mapping(finat_element, phys_element)
+
+
+def check_zany_piola_mapping(finat_element, phys_element):
```
rckirby, line 110: I approve of this.  We kept doing something like this over and over.

test/test_zany_mapping.py
```diff
@@ -100,3 +105,104 @@ def test_argyris_point(ref_cell, phys_cell):
...
+    Mh = np.linalg.solve(Phi @ Phi.T, Phi @ phi.T).T
+    M = M[:num_facet_dofs]
+    Mh = Mh[indices][:num_facet_dofs]
+    Mh[abs(Mh) < 1E-10] = 0.0
+    M[abs(M) < 1E-10] = 0.0
+    assert np.allclose(M, Mh), str(M.T - Mh.T)
+    assert np.allclose(ref_vals_zany[:num_facet_dofs], phys_vals[indices][:num_facet_dofs])
+
+
+@pytest.mark.parametrize("element", [
```
rckirby, line 166: cleaner/shorter if we import all the elements from finat and then save a few lines on the parameter list?

rckirby (changes requested):
Good work on this.  You found some nice opportunities to streamline things, let's make sure to test  Stokes stability/convergence in Firedrake before we merge and fix (or disable?) anything that breaks.

finat/alfeld_sorokina.py
```diff
@@ -0,0 +1,42 @@
...
+
+from finat.fiat_elements import FiatElement
+from finat.physically_mapped import Citations, PhysicallyMappedElement
+from finat.piola_mapped import piola_inverse
+
+
+class AlfeldSorokina(PhysicallyMappedElement, FiatElement):
+    def __init__(self, cell, degree=2):
+        if degree != 2:
+            raise ValueError("Alfeld-Sorokina only defined for degree = 2")
```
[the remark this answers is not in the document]
pbrubeck: The space can be generalized to higher degree quite easily, we don't have the higher order DOFs yet.

finat/arnold_qin.py
```diff
@@ -0,0 +1,70 @@
...
+
+        for f in sorted(dofs[sd-1]):
+            rows = numpy.asarray(transform(fiat_cell, J, detJ, f))
+            fdofs = dofs[sd-1][f]
+            bfs = edofs[sd-1][f][1:]
+            V[numpy.ix_(bfs, fdofs)] = rows[..., :len(fdofs)]
+        return ListTensor(V.T)
+
+
+class ReducedArnoldQin(ArnoldQin):
```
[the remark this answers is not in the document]
pbrubeck: Both the full and reduced elements  make sense for Stokes and are going to become available in Firedrake.

finat/piola_mapped.py
pbrubeck: Even in 3D, we only call this code for 2 x 2 matrices.

finat/alfeld_sorokina.py
```diff
@@ -0,0 +1,42 @@
...
+        if Citations is not None:
+            Citations().register("AlfeldSorokina2016")
+        super().__init__(FIAT.AlfeldSorokina(cell, degree))
+
+    def basis_transformation(self, coordinate_mapping):
+        sd = self.cell.get_spatial_dimension()
+        bary, = self.cell.make_points(sd, 0, sd+1)
+        J = coordinate_mapping.jacobian_at(bary)
+        detJ = coordinate_mapping.detJ_at(bary)
+        adjJ = piola_inverse(self.cell, J, detJ)
```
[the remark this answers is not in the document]
pbrubeck: It stands the adjugate, but we can hide this from the name, as it is irrelevant.

finat/physically_mapped.py
rckirby: This would go in `/tsfc/gem/gem.py`.  Note: there is already an `Inverse` node in GEM there.  We'd have to track down how it gets lowered.

pbrubeck: We will address properly on a separate TSFC PR

finat/alfeld_sorokina.py
rckirby, line 13: Isn't `NotImplementedError` better if it's defined but we haven't got it working yet?

rckirby (approved):
I'm approving, just one minor quibble on `ValueError` vs `NotImplementedError`

rckirby (approved):

PR #141, opened 2024-10-14
Fix intersphinx mapping

pbrubeck:


dham (approved):

docs/source/conf.py
```diff
@@ -289,5 +289,9 @@
 # texinfo_no_detailmenu = False
 
 
-# Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {"https://docs.python.org/3/": None}
+# Configuration for intersphinx
+intersphinx_mapping = {
+    'FIAT': ('https://fenics.readthedocs.io/projects/fiat/en/latest/', None),
```
wence-, line 294: Given that firedrakeproject/fiat is where active fiat development lives, should the docs be published from there somehow, and the intersphinx mapping updated? This link is to an out-of-date version, no?

I suppose that is a bigger modification than just adding a url here.

pbrubeck: Right, this is pointing to the docs for the fenics master branch. I got this link from the firedrake docs/source/conf.py script, so it is  also out-of-date there.

dham (changes requested):
For goodness sake check that this actually works.

docs/source/conf.py
connorjward: Please make an issue stating this.

PR #143, opened 2024-10-28
Cleanup H(div, S) elements

pbrubeck:


finat/aw.py
```diff
@@ -90,16 +83,9 @@ def entity_dofs(self):
                 1: {0: [0, 1, 2, 3], 1: [4, 5, 6, 7], 2: [8, 9, 10, 11]},
                 2: {0: [12, 13, 14]}}
 
-    def entity_closure_dofs(self):
```
rckirby, line 93: I thought this override was needed since we are changing the number of degrees of freedom relative to the FIAT element and can't just pass it through?  I forget (it's been years) but there was something somewhere that broke without this.

rckirby (approved):
Mostly fine -- I thought we needed `entity_closure_dofs` intact, but if it's not breaking anything in the firedrake stack maybe we don't need it any more?

